PR #1, opened 2021-07-21
DDO-1362 Refactor CI/CD to build 2 images which are versioned together

mflinn-broad:
This refactors the ci/cd pipeline from the [go template repo](https://github.com/broadinstitute/golang-project-template) to support building two binaries from this repo and packaging both into separate docker images. 

This is so that the CLI and server can be versioned together and helps to ensure that build pipelines will always run the correct version of the client against the server.

The two dockerfiles will likely need to diverge at least slightly down the road, ie different `ENTRYPOINT`. I am keeping them largely the same for now for the sake of simplicity.

[Here](https://console.cloud.google.com/artifacts/docker/dsp-artifact-registry/us-central1/sherlock?project=dsp-artifact-registry) is how this ends up looking in artifact registry.


choover-broad (approved):

go.mod
```diff
@@ -1,3 +1,3 @@
-module github.com/broadinstitute/golang-project-template
+module github.com/broadinstitute/sherlock
 
 go 1.15
```
jyang-broad, line 3: el quicky, go version mismatch.

mflinn-broad: el good catch 😄 I should probably remove go.mod from the template repo. That should really be generated per project.

PR #3, opened 2021-08-02
DDO-1363 register services and associated builds

mflinn-broad:
This is part one of a 2 part ticket to enable sherlock to be able to both list the existing `services` it tracks and `POST` new services. 

This first pr implements support for just listing existing services. It also includes some helper methods for seeding the database in integration tests.

Additionally it also adds support for introducing separation between unit and integration tests and being able to run both  locally and in CI/CD. They can be run locally with `make unit-test` and `make integration-test` respectively. 

The diff looks a bit scary but it's mostly tweaks to dockerfiles to support integration tests and cleaning out some unneeded boilerplate from the template repo.

*Now also switched over to gin and gorm*


cmd/server/main.go
```diff
@@ -1,69 +1,44 @@
...
 
 	"log"
 
-	"github.com/golang-migrate/migrate/v4"
+	"github.com/broadinstitute/sherlock/internal/db"
+	"github.com/broadinstitute/sherlock/internal/sherlock"
+	"github.com/broadinstitute/sherlock/internal/tools"
 	_ "github.com/golang-migrate/migrate/v4/database/postgres"
 	_ "github.com/golang-migrate/migrate/v4/source/file"
 	_ "github.com/jackc/pgx/v4/stdlib"
```
jack-r-warren, line 14: This is a DB driver, right? Is it a problem that we're using a different driver here versus in the migrations file?

internal/services/services.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+
+	"github.com/broadinstitute/sherlock/internal/db"
+)
+
+// Service is the data structure representing an indvidual applicaiton
+type Service struct {
+	ID        int    `json:"id,omitempty" db:"id"`
```
jack-r-warren, line 11: Can this ever be empty?

jack-r-warren (approved):

internal/sherlock/routes.go
```diff
@@ -0,0 +1,34 @@
+package sherlock
+
+import (
```
choover-broad, line 3: Similarly, I'm wondering if we should look at REST frameworks for Golang, I bet there are some good options out there

choover-broad (commented):
Left a few comments, will loop back for a deeper dive later 👍 

Love all the up-front thought about developer experience and integration testing!

internal/services/services.go
mflinn-broad: When POSTing a new service yes, when listing existing ones no.

cmd/server/main.go
mflinn-broad: Whoops good catch! I was originally using the same driver for both then noticed it was deprecated in favor of `pgx` which is what was used here. Forgot to update the migrations though, will fix.

mflinn-broad: fixed!

internal/sherlock/routes.go
mflinn-broad: 😄

internal/sherlock/sherlock.go
```diff
@@ -0,0 +1,62 @@
...
+	"log"
+	"net/http"
+
+	"github.com/broadinstitute/sherlock/internal/db"
+	"github.com/spf13/viper"
+)
+
+// package level variable holding a viper instance that will manage sherlock's config
+var (
+	Config = viper.New()
```
jack-r-warren, line 13: Just curious, why is viper being wrapped?

internal/sherlock/routes.go
```diff
@@ -0,0 +1,30 @@
...
+	"net/http"
+
+	"github.com/broadinstitute/sherlock/internal/services"
+	"github.com/gin-gonic/gin"
+)
+
+// helper function to take an existing sherlock application instance
+// then build and attach a gin router to it.
+// this decouples building the router from instanting a sherlock Application
+// which makes testing easeier
```
jack-r-warren, line 13:
```suggestion
// which makes testing easier
```

internal/sherlock/sherlock_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/broadinstitute/sherlock/internal/services"
+	"github.com/broadinstitute/sherlock/internal/sherlock"
+	"github.com/broadinstitute/sherlock/internal/tools"
+	"github.com/google/go-cmp/cmp"
+)
+
+// expopses a common sherlock instance that can be shared in integration tests
+var app *sherlock.Application
+
+// This integration test patter is taken from https://www.ardanlabs.com/blog/2019/10/integration-testing-in-go-set-up-and-writing-tests.html
```
jack-r-warren, line 19:
```suggestion
// This integration test pattern is taken from https://www.ardanlabs.com/blog/2019/10/integration-testing-in-go-set-up-and-writing-tests.html
```

jack-r-warren (approved):
This looks amazing, literally the only thing I found were a few comment typos. Awesome stuff

internal/sherlock/sherlock.go
mflinn-broad: It's not really being wrapped unless I'm misunderstanding you. It's storing a viper instance as a package level variable. This is mainly done so that all the config initialization can happen in the sherlock package `init` function, which in turn is guaranteed to run before `main`. This is mainly to ensure the config is fully read in and parsed before anything will try to access it.

internal/sherlock/sherlock_integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+
+	"github.com/broadinstitute/sherlock/internal/db"
+	"github.com/broadinstitute/sherlock/internal/services"
+	"github.com/broadinstitute/sherlock/internal/sherlock"
+	"github.com/broadinstitute/sherlock/internal/tools"
+	"github.com/google/go-cmp/cmp"
+)
+
+// expopses a common sherlock instance that can be shared in integration tests
```
jyang-broad, line 16: s/expopses/exposes

internal/tools/seed.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	err = repository.DB.Find(&services).Error
+	return services, err
+}
+
+// Truncate cleans up tables after integration tests
+func Truncate(repository *db.Repository) error {
+	// gorm doesn't seem to support truncate operations which are essential to cleaning up after
+	// integration tests (and the only use case of this function so doing it with raw sql)
+	truncateStatement := "TRUNCATE TABLE services, builds, environments, service_instances, deploys"
```
jyang-broad, line 38: looks sane! Sad we have to manually do it.

jyang-broad (approved):
Looks sane! Just a couple readability/typo comments.

internal/db/db.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := m.Up(); err != nil {
+		return err
+	}
+
+	log.Println("database migration complete")
+	return nil
+}
+
+func Connect(config *viper.Viper) (*gorm.DB, error) {
+	dsn := fmt.Sprintf(
```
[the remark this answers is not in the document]
mflinn-broad: That was me copy pasting from gorm's usage docs, happy to change it to something more clear.

internal/tools/seed.go
```diff
@@ -0,0 +1,42 @@
+package tools
+
+import (
+	"github.com/broadinstitute/sherlock/internal/db"
+	"github.com/broadinstitute/sherlock/internal/services"
+)
+
+// SeedServices is a test utility that will populate a database with a predetermined list of "services"
+// to be used for running integration tests against a real database
+func SeedServices(repository *db.Repository) ([]services.Service, error) {
```
choover-broad, line 10: I see a couple places where the `Repository`/database client is passed in as a parameter where I'd either expect either (a) the function to use a `Repository` as a receiver or (b) the `Repository`/database to follow the singleton pattern, kept in global state* 🤔 This parameter passing seems a bit verbose coming from Ruby.

*I know this can be controversial haha. This blog post was pretty interesting: https://alexedwards.net/blog/organising-database-access

choover-broad (approved):
This looks great! Left some thoughts/comments in-line which you can feel free to ignore. 😁 I really, really, appreciate all the focus on code design, testing, CI/CD, and developer experience upfront. And it's very exciting to see a proof-of-concept endpoint! 🎉💯

internal/tools/seed.go
choover-broad: And I know the "BUT THIS IS HOW RUBY DOES ITTTT" comments must be getting old 😁

internal/db/db.go
```diff
@@ -0,0 +1,104 @@
...
+	m, err := migrate.NewWithDatabaseInstance(
+		changelogLocation,
+		databaseName,
+		driver,
+	)
+	if err != nil {
+		return err
+	}
+
+	if err := m.Up(); err != nil {
```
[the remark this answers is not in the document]
mflinn-broad: So there is some go stuff at play here. `migrate` is actually the name space identifier for the imported `golang-migrate` library. These namespace identifiers by default are the final element that isn't a version identifier in the import url. so in this case methods and types imported like this `import github.com/golang-migrate/migrate/v4` get used like `migrate.MethodFromMigrateLibrary()`. 

the m variable was me copying from usage docs again, I can change it to something else.

cmd/server/main.go
```diff
@@ -1,69 +1,32 @@
...
-		os.Exit(1)
-	}
-	defer db.Close()
-
-	log.Println("Successfully connected to database")
 
 	// just logging error rather than failing here as
 	// a no changes migration is treated as an error..
-	if err := applyMigrations(); err != nil {
+	if err := db.ApplyMigrations("db/migrations", sherlock.Config); err != nil {
```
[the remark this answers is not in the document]
mflinn-broad: I think eventually we don't want to run the migrations in app at all and use something like Jack's liquibase migration job.

I only really see this being enabled in local  development and integration test workflows. I can make it fail if you feel strongly though. I mainly had it as a pass through because the failures were annoying me when I was trying to fix things :)

internal/db/db.go
```diff
@@ -0,0 +1,104 @@
...
+	// indirect import used to set proper migration data source
+	_ "github.com/golang-migrate/migrate/v4/source/file"
+	_ "github.com/jackc/pgx/v4/stdlib"
+)
+
+const databaseName string = "sherlock"
+
+// Repository is a wrapper around a connection to a data base
+//  that can support mocking database connections in tests
+type Repository struct {
```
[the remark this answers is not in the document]
mflinn-broad: this ties into your final comment, will respond in more detail there.

internal/db/db.go
```diff
@@ -0,0 +1,104 @@
...
+func NewRepository(db *gorm.DB) *Repository {
+	return &Repository{DB: db}
+}
+
+// ApplyMigrations is a utility function intended for use in integration tests and
+// local development where changelogs can be applied to a local postgres instance
+// during startup
+func ApplyMigrations(changeLogPath string, config *viper.Viper) error {
+	// check for environment flag whether to run migrations on app start up or not
+	if _, ok := os.LookupEnv("SHERLOCK_INIT_DB"); !ok {
```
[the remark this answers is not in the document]
mflinn-broad: yup good call

internal/db/db.go
```diff
@@ -0,0 +1,104 @@
...
+		config.GetString("dbport"),
+		config.GetString("dbname"),
+		config.GetString("dbssl"),
+	)
+
+	log.Println("Executing database migration")
+	// The below code is to ensure migrations run using the same
+	// postgres driver (pgx) that gorm uses. golang-migrate uses
+	// a different postgres driver by default
+	db, err := sql.Open("pgx", dbURL)
```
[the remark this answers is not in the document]
mflinn-broad: yup this set off similar alarm bells for me. Then I kind of moved on in the interest of continuing to make progress. I think what you described *might* be possible. I'll give it a quick shot!

internal/tools/seed.go
mflinn-broad: Thank you for calling me out on this. I was actually basing my database access patterns on section 3 from that very article. However I did forget to go back and turn the methods that accept because I got distracted by the gin/gorm stuff  so thank you for bringing attention to this. 

There is one go quirk that comes into play here. You can only define receiver methods in the same `package` where the receiver type is defined. So that would not be doable in the case of these seeding helper functions. 

I would like to keep some separation with these specific methods as they are only utilities for integration test setup and not part of core application logic. So I think this particular file should stay as it is. 

I will go back and update the other functions in the `db` package that use this same pattern to take a receiver instead though :).

mflinn-broad: Also agree `repository` isn't the best name for the type. That was just a convention I saw online and was having trouble thinking of a good alternative..

internal/db/db.go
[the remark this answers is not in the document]
jyang-broad: ah got it. I got confused because I couldn't figure out where `migrate` was coming from.

internal/tools/seed.go
choover-broad: > You can only define receiver methods in the same package where the receiver type is defined. 
Ahhh... TIL! In retrospect this makes a lot of sense.

Thanks for the clarification! 👍

PR #7, opened 2021-09-07
DDO-1493 DDO-1493 Refactor for better decoupling/extensibility

mflinn-broad:
The goal of this pr is to perform a reorganization of the code structure of sherloc's `/services` api end point group, with the idea that this organization structure will serve as a model for future more complex endpoint groups such as `/builds` and `/deploys. 

There are two primary goals in this refactor. 

1. Decouple the types used for modeling entites in the database from types used reading and writing responses to and from clients. 
2. Decouple various layers in the system so that they are easily modifiable and testable. The goal here is to introduce separation of concerns between routing, middleware /business logic, and database /persistence layers in the code. This is achieved through the use of go's `interface` types. This will make routing behavior, business logic and database operations all unit testable separately and avoid dependencies in multiple parts of the system that require updates in expected locations when making code changes.

We also no longer need an extremely complex sql mocking library to unit test routing and validation or other business logic.


jack-r-warren (approved):
I actually pulled all of the changes into GoLand locally, and this looks great

The one comment I have is that I have is that we'll need to make sure we come up with some good documentation mechanism for the API. The way we have it set up now, there's no single file with all the routes. I think that makes sense from a code standpoint, just means we can't say "we don't have a swagger page but you can look at this one file for all the routes"

mflinn-broad:
> I actually pulled all of the changes into GoLand locally, and this looks great
> 
> The one comment I have is that I have is that we'll need to make sure we come up with some good documentation mechanism for the API. The way we have it set up now, there's no single file with all the routes. I think that makes sense from a code standpoint, just means we can't say "we don't have a swagger page but you can look at this one file for all the routes"

Great point. I think there is a tool we can use to code-gen the swagger page even if the routes are spread across multiple files but agreed with your point. This is something I need to keep in mind and look into a little more.

PR #8, opened 2021-09-13
DDO-1401 API Group For Managing Build Entities

mflinn-broad:
This pr implements the minimal functionality needed for sherlock to track builds reported from various ci/cd systems. 
There are 3 core API functionalities implemented in this pr.

1. Listing out all builds via `GET /builds`
2. Retrieving a specific build by ID via `GET /builds/:id`
3. Creating a new build entity via `POST /builds`

There is a scary looking diff here but a good chunk if is actually just various cases for table driven tests.
Some off the test code is setting off some DRY alarm bells but it works fine. Going to create a separate ticket to do some test cleanup.

In addition to the main functionality implemented above this pr includes some minor refactors to the Public methods for `ServiceController` to avoid some potential separation of concerns issues that came up when implementing the functionality to create a new service entity when sherlock receives a create build request for a service that doesn't currently exist in the database.


internal/builds/builds.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+// Response is a type that allows all data returned from the /builds api group to share a consistent structure
+type Response struct {
+	Builds []Build `json:"builds"`
+	Error  string  `json:"error,omitempty"`
+}
+
+// CreateBuildRequest is the type used to validate a request for a new build
```
choover-broad, line 37: Should there be a separate struct type here for serializing Builds in responses? I realize it adds overhead but it's a bit surprising that fields in requests are snake case and fields in responses are camel case. And I think it will be useful later on (for hiding model details in the API, maintaining backwards compatibility, populating some fields w/ code, etc).

jack-r-warren (approved):
Looks good to me, fan of the ID -> Name refactoring

internal/sherlock/sherlock_integration_test.go
```diff
@@ -174,6 +179,201 @@ func Test_sherlockServerIntegration(t *testing.T) {
...
+		var result builds.Response
+		if err := json.NewDecoder(response.Body).Decode(&result); err != nil {
+			t.Errorf("error decoding response body: %v", err)
+		}
+
+		assert.Equal(t, 1, len(result.Builds))
+		assert.Empty(t, result.Error)
+		assert.Equal(t, newBuildRequest.VersionString, result.Builds[0].VersionString)
+	})
+	t.Run("POST /builds with new service", func(t *testing.T) {
```
choover-broad, line 268: Love all this integration testing!

Can we add some coverage for a version conflict, i.e. client tries to publish a new build with a version string that already exists? Curious how user-friendly the error message is

choover-broad (approved):
Exciting progress! Didn't have time to get super-detailed, left a few comments in-line

internal/builds/builds.go
mflinn-broad: great point. I think I will tackle this in a separate ticket though!

internal/sherlock/sherlock_integration_test.go
mflinn-broad: yep yep, will do

mflinn-broad: I think currently it bubbles up an error from the DB itself saying that a `unique` constraint was violated. Could definitely wrap that in something more user friendly

mflinn-broad: All set https://github.com/broadinstitute/sherlock/pull/8/commits/d65ae979a2c2b235277b5154a263e787ed61a999

internal/builds/builds.go
mflinn-broad: [DDO-1532](https://broadworkbench.atlassian.net/browse/DDO-1532)

PR #9, opened 2021-09-15
DDO-1532 Serializers for service and build responses

mflinn-broad:
This pr was inspired by @choover-broad 's comment [here](https://github.com/broadinstitute/sherlock/pull/8#discussion_r708553219). The goal here is to create serializers to convert the model type for Sherlock's entities to a type that is used generate http response bodies. 

This means that the DB model type is now fully decoupled from requests and responses. This is important for giving us more granular control over what information we show in these api responses. This also makes it easier to hide implementation details and support backwards compatibility if needed.

These changes are based on [this pattern](https://github.com/gothinkster/golang-gin-realworld-example-app/blob/master/articles/serializers.go) from the gin example repo.

### DIFF
--- a/internal/builds/serializers.go
+++ b/internal/builds/serializers.go
@@ -1,0 +1,51 @@
+package builds
+
+import (
+	"time"
+
+	"github.com/broadinstitute/sherlock/internal/services"
+)
+
+// Response is a type that allows all data returned from the /builds api group to share a consistent structure
+type Response struct {
+	Builds []BuildResponse `json:"builds"`
+	Error  string          `json:"error,omitempty"`
+}
+
+type BuildResponse struct {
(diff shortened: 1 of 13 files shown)


choover-broad (approved):
Nice! Thanks, this looks really clean 🥇

jyang-broad (approved):
Look Sane!

internal/builds/serializers.go
```diff
@@ -0,0 +1,57 @@
+package builds
```
jyang-broad, line 1: Can you add some file-level comments? I'm trying to add these to the files I have so it's easier to sort what type of logic should belong or shouldn't belong in a file.

mflinn-broad: ✅

PR #12, opened 2021-09-21
[No Ticket] Fix Bug in BuiltAt timestamp field

mflinn-broad:
In gorm models `CreatedAt` and `UpdatedAt` fields have special magic going on behind the scenes where it will set those timestamps to the current time in postgres unless they are specified ie non-zero value for `time.Time` . To get this same behavior for other timestamp columns such as built at requires a special `gorm` struct tag. This will guarantee that `built_at` column is set to current time when a new build is created if it is not otherwise specified. 

The behavior without this is that a Zero value for go's `time.Time` would be persisted to Postgres in the `built_at` column which can cause non-deterministic results. [Relevant documentation](https://gorm.io/docs/models.html#Creating-Updating-Time-Unix-Milli-Nano-Seconds-Tracking)

### DIFF
--- a/internal/tools/seed.go
+++ b/internal/tools/seed.go
@@ -2,6 +2,7 @@ package tools
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/broadinstitute/sherlock/internal/builds"
 	"github.com/broadinstitute/sherlock/internal/services"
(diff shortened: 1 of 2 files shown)


internal/tools/seed.go
```diff
@@ -78,6 +79,7 @@ func SeedBuilds(db *gorm.DB) ([]builds.Build, error) {
 			VersionString: "gcr.io/workspacemanager:1.2.0",
 			CommitSha:     "6a5s4df",
 			BuildURL:      "https://build.3.log",
+			BuiltAt:       time.Now().Add(-6 * time.Hour),
```
jyang-broad, line 82: Assuming the -6 is to put the time into the past, this is fine (but I guess could use a comment). I thought for a sec this was to adjust for timezone but I'm pretty sure that's not the case (cuz it wouldn't work).

knapii-developments (approved):

internal/tools/seed.go
mflinn-broad: good call ✅

PR #14, opened 2021-09-24
DDO-1562 Service Instance Functionality and Integration Test Refactors

mflinn-broad:
This PR adds some basic functionality for `serviceInstances`. We can now seed the db with mock service instance data for testing and list out all `serviceInstances`. I put the model and controller code for `serviceInstances` as part of `package deploys` as they are essentially an intermediary abstraction between services environments and a deploy. 

AFAIK the only functionality that cares about them are `deploys` so I included in package deploys in order to group associated functionality together. Open to counter arguments / other ideas about where to put this though.

Additionally this also performs some refactors of the existing integration testing patterns. I've applied the refactors to `serviceInstance_test.go` and @jyang-broad 's `environment_test.go`. Holding off on applying to  / breaking up `sherlock_integration_test.go` 

[Here](https://github.com/broadinstitute/sherlock/blob/1840ae2a58f4e8413dd4e1ef5f0f8fb1537c04db/internal/deploys/serviceInstance_test.go#L13) and [here](https://github.com/broadinstitute/sherlock/blob/1840ae2a58f4e8413dd4e1ef5f0f8fb1537c04db/internal/environments/environments_test.go#L38) are examples of what integration tests could look like post refactor.


em-may (approved):
Not a huge fan of raw SQL, but if you have to then I guess so...

mflinn-broad:
> Not a huge fan of raw SQL, but if you have to then I guess so...

I assume you are talking about [this](https://github.com/broadinstitute/sherlock/blob/cd054e971cd80bce70c7c381d52d5d470865bda3/internal/testutils/integrationHelpers.go#L59). That is literally the only line of raw sql in the entire application. It only gets used in testing.

PR #16, opened 2021-09-28
Ddo 1521 environment integrations tests

jyang-broad:
POC PR, still needs some more tests to be fully tested,
There are 2 types of tests, one is using a "testCases" format (CreateEnvironment), the other is using a more freeform format, not sure what's better atm.

### DIFF
--- a/internal/environments/environments_test.go
+++ b/internal/environments/environments_test.go
@@ -59,7 +59,8 @@ func initTestApp(t *testing.T) *TestApplication {
 		Environments: NewController(dbConn),
 		db:           dbConn,
 	}
-	testutils.Truncate(app.db)
+
+	testutils.BeforeIntegrationSubTestCleanup(t, app.db)
 
 	return app
 }
@@ -136,7 +137,8 @@ func (suite *EnvironmentTestSuite) TestIntegrationCreateEnvironments() {
 
 			// create all non-final environments as setup
 			for _, request := range testCase.requests[:len(testCase.requests)-1] {
-				suite.testApp.Environments.CreateNew(request)
+				_, err := suite.testApp.Environments.CreateNew(request)
+				assert.Equal(suite.T(), nil, err)
 			}
 
 			// create and test the last environment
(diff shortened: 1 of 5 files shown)


internal/environments/environments_test.go
```diff
@@ -3,68 +3,230 @@
...
-type TestApplication struct {
-	Environments *EnvironmentController
-	db           *gorm.DB
+		suite.testApp.Environments.CreateNew(suite.goodEnvironmentRequest)
+
+		foundEnvironment, err := suite.testApp.Environments.GetByName(suite.goodEnvironmentRequest.Name)
+
+		assert.Equal(suite.T(), foundEnvironment.Name, suite.goodEnvironmentRequest.Name)
+
+		assert.Equal(suite.T(), nil, err)
```
mflinn-broad, line 163: you can also use `assert.NoError(t, err)` to do this

internal/environments/environments_test.go
```diff
@@ -3,68 +3,230 @@
...
+		testutils.BeforeIntegrationSubTestCleanup(suite.T(), suite.testApp.db)
 
-	return app
+		foundEnvironments, err := suite.testApp.Environments.ListAll()
+
+		assert.Equal(suite.T(), len(foundEnvironments), 0)
+		assert.Equal(suite.T(), nil, err)
+	})
+
+	suite.Run("ListAll returns one Environment", func() {
```
mflinn-broad, line 189: This is probably a better way to handle listAll integration tests than the complicated seeding I was trying to do :)

internal/testutils/integrationHelpers.go
```diff
@@ -71,3 +79,10 @@ func Cleanup(t *testing.T, dbConn *gorm.DB) {
 		t.Fatalf("error cleaning db after test run: %v", err)
 	}
 }
+
+// Cleanup intended to be called in between subtests
+func BeforeIntegrationSubTestCleanup(t *testing.T, dbConn *gorm.DB) {
```
mflinn-broad, line 84: Can we just call this cleanup? it isn't necessarily specific to subtests

internal/testutils/seed.go
```diff
@@ -0,0 +1,96 @@
...
+	"time"
+
+	"github.com/broadinstitute/sherlock/internal/builds"
+	"github.com/broadinstitute/sherlock/internal/services"
+	"gorm.io/gorm"
+)
+
+// SeedServices is a test utility that will populate a database with a predetermined list of "services"
+// to be used for running integration tests against a real database
+func SeedServices(db *gorm.DB) ([]services.Service, error) {
```
mflinn-broad, line 14: Can we keep the seeding functions in their respective pacakages rather than `testutils`. This will cause import cycles if I wanted to use other methods from `package testutils` from an integration test in `package services`

mflinn-broad (approved):
Testing code looks solid to me. Just a couple small comments. Big one is where should we put seeding functions.

internal/testutils/integrationHelpers.go
jyang-broad: probably, this was going to be specific to subtests, but then I just ended up using Cleanup

mflinn-broad (approved):
lgtm

PR #17, opened 2021-09-29
Mitigating Race Condition issue in Github Actions Workflow

mflinn-broad:
This  PR aims to address the race condition issue that I mentioned in standup that started popping up in the sherlock integration test suite. Part of what made this issue so difficult to debug was that go's testing tool was being invoked with different configuration in local environments vs the gitub actions pipeline.

Now the integration tests will be executed in a consistent manner both in GHA and on developer machines.

Additionally my initial experimental approach to integration testing `sherlock_integration_testing.go` has outgrown it's original scope as it has proven to be an untenable pattern for testing as the application grows in scope and complexity and is now leaking side effects into other tests. 

As such I'm commenting out that file for now and will break up testing the same functionality into separate files using the same pattern @jyang-broad and I have started adopting for integration testing [here](https://github.com/broadinstitute/sherlock/blob/main/internal/environments/environments_test.go) using `testify/suite`. This is a much more scalable approach and should be what integration testing in sherlock looks like going forwards. 

The refactoring of `sherlock_integration_test.go` into separate `suites` will be handled in a separate pr.

### DIFF
--- a/build/local/server/docker-compose.test.yaml
+++ b/build/local/server/docker-compose.test.yaml
@@ -18,6 +18,8 @@ services:
       - database
     networks: 
       - sherlock-local-dev
+    volumes:
+      - $PWD:/build
 
   database:
     container_name: postgres
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -17,20 +17,6 @@ jobs:
   # Integration tests are handled in a separate job
   test:
     runs-on: ubuntu-latest
-    services:
-      postgres:
-        image: postgres:13
-        # Not real db credentials, only used for ephemeral integration test runs
-        env:
-          POSTGRES_USER: sherlock
-          POSTGRES_PASSWORD: password
-        options: >-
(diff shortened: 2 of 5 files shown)


build/local/server/docker-compose.test.yaml
```diff
@@ -18,6 +18,8 @@ services:
       - database
     networks: 
       - sherlock-local-dev
+    volumes:
+      - $PWD:/build
```
jyang-broad, line 22: I know we need the build folder, not actually sure why we need it in volumes, but I'm just assuming we do now.

jyang-broad (approved):
looks sane! Also, i ran make tests-with-coverage locally and it worked so yay!

build/local/server/docker-compose.test.yaml
mflinn-broad: it's so the coverage report is accessible to the codecov action step in the pipeline. Otherwise we'd loose it when the container stops.

PR #18, opened 2021-09-30
DDO-1581 Fixing bad Integration Testing patterns in favor of testify/suite

mflinn-broad:
This pr aims to remove the original poc database integration testing patterns I implemented for sherlock. While those patterns served their original purpose they have become slow and unreliable as the application has grown in size and complexity. 

This pr does 3 main things:

1. Remove the old unwieldy `sherlock_integration_test.go` file. 
2. Cover the functionality previously covered by ^ with separate tests in the respective packages using the `testify/suite` testing patterns
3. Remove the docker compose based integration test tooling in favor of just spinning up a postgres container and running `go test` locally against it.

The 3rd point is a major quality of life improvement as the docker-compose integration testing pattern was slow and unreliable and difficult to debug. With this change the time to run the full tests suite has decreased from 1-2 minutes to 2-3 seconds.


makefile
```diff
@@ -7,16 +7,22 @@ local-stop:
...
 integration-test:
-	docker-compose -f build/local/server/docker-compose.test.yaml up --build --abort-on-container-exit
-	docker-compose -f build/local/server/docker-compose.test.yaml down --volumes
-	
-integration-down:
-	docker-compose -f build/local/server/docker-compose.test.yaml down --volumes
+	export SHERLOCK_DBHOST="localhost"
+	export SHERLOCK_DBPASSWORD="password"
+	docker run --name test-postgres -e POSTGRES_PASSWORD=password -e POSTGRES_USER=sherlock -d -p 5432:5432 postgres:13
+	go test -v -race ./...
```
knapii-developments, line 15: Question for learning:
What's the difference between this line and line 20. I see here the arguments are different and it seems to run in a docker container. I'm don't understand why it's under unit-test as well?

internal/builds/builds_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		newBuild := CreateBuildRequest{}
+
+		_, err := suite.app.builds.CreateNew(newBuild)
+
+		suite.Require().Error(err)
+	})
+}
+
+func (suite *BuildsIntegrationTestSuite) TestGetByID() {
```
jyang-broad, line 60: Probably a better naming convention that I've been using 👍

makefile
mflinn-broad: The test command itself doesn't run in docker here. It runs a postgres container that the tests can then connect to so they can perform real database operations. The main difference between this and line 20 is the use of the `-short` flag which tells the testing tool to skip any tests that require external dependencies such as postgres. 

So if you were to run `make integration-test` in this repo it would run the full test suite against a real postgres instance. If you ran `make unit-test` it would only run the unit tests (those with no external dependencies)

internal/builds/builds_test.go
```diff
@@ -0,0 +1,102 @@
...
+	suite.Run(t, new(BuildsIntegrationTestSuite))
+}
+
+func (suite *BuildsIntegrationTestSuite) SetupSuite() {
+	suite.app = initTestApp(suite.T())
+	// ensure the db is clean before running suite
+	testutils.Cleanup(suite.T(), suite.app.db)
+}
+
+func (suite *BuildsIntegrationTestSuite) TestCreateBuild() {
```
jyang-broad, line 30: Request: add uniqueness test on version_string

internal/services/models.go
```diff
@@ -24,8 +24,8 @@ type dataStore struct {
 
 // Service is the data structure that models a service entity persisted to a database via gorm
 type Service struct {
-	ID        int
-	Name      string
+	ID        int    `gorm:"primaryKey" faker:"unique"`
+	Name      string `gorm:"not null;default:null"`
```
jack-r-warren, line 28: Shame that gorm doesn't have a more semantic "required" thing, at least their doc page is good

jack-r-warren (approved):
This looks great, just dreading replicating these patterns in Revere lol

internal/services/services_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+		suite.Assert().Equal(len(services), 0)
+		suite.Assert().NoError(err)
+	})
+
+	suite.Run("ListAll returns one Environment", func() {
+		testutils.Cleanup(suite.T(), suite.app.db)
+
+		newService := CreateServiceRequest{
+			Name:    faker.Name(),
```
jyang-broad, line 74: nice! randomized inputs!

internal/services/services_test.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+
+			_, err := suite.app.services.CreateNew(newService)
+			suite.Assert().NoError(err)
+		}
+
+		services, err := suite.app.services.ListAll()
+		suite.Require().NoError(err)
+
+		suite.Assert().Greater(len(services), 1)
```
jyang-broad, line 105: I think `Equal(len(services),3)` is cleaner, unless we have a reason to expect a random number greater than 1.

internal/services/services_test.go
```diff
@@ -0,0 +1,161 @@
...
+		suite.Assert().Equal(createdService.ID, id)
+	})
+
+	suite.Run("errors on non-existent service", func() {
+		testutils.Cleanup(suite.T(), suite.app.db)
+
+		_, err := suite.app.services.GetByName("tester")
+		suite.Assert().ErrorIs(err, ErrServiceNotFound)
+
+		_, ok := suite.app.services.DoesServiceExist("tester")
```
jyang-broad, line 137: not really test related, but reading this, `doesExist` feels like a better var name. I actually just looked up the method to figure out what `ok` means.

internal/testutils/integrationHelpers.go
```diff
@@ -66,8 +66,19 @@ func ConnectAndMigrate(t *testing.T) *gorm.DB {
 func Truncate(db *gorm.DB) error {
 	// gorm doesn't seem to support truncate operations which are essential to cleaning up after
 	// integration tests (and the only use case of this function so doing it with raw sql)
-	truncateStatement := "TRUNCATE TABLE services, builds, environments, service_instances, deploys, clusters, allocation_pools"
-	err := db.Exec(truncateStatement).Error
+	testDeleteStatement := `
```
jyang-broad, line 69: What's the reason to move to DELETE from TRUNCATE? I honestly have no idea what the difference is so just curious.

jyang-broad (approved):
Looks sane! Really like how the new code reads and also the faker random inputs.

internal/builds/builds_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	suite.Run("fails with empty create request", func() {
+		testutils.Cleanup(suite.T(), suite.app.db)
+
+		newBuild := CreateBuildRequest{}
+
+		_, err := suite.app.builds.CreateNew(newBuild)
+
+		suite.Require().Error(err)
```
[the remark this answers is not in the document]
mflinn-broad: Require will fail the test immediately. I'm using require mainly to check for things that can go wrong that would make the rest of the test guaranteed to fail. Basically if an error shows up that is going to guarantee the rest of the test fails II check it with require. Cuts down on error noise in the test output imo.

internal/builds/builds_test.go
mflinn-broad: nods, good catch

internal/services/services_test.go
mflinn-broad: This was me trying to hack around the non-deterministic database issues I was having. Good call I'll change it back

internal/services/services_test.go
mflinn-broad: sure I'm happy to change it. The `ok` thing was just I convention I saw in other go code bases. Not attached to it though

internal/testutils/integrationHelpers.go
mflinn-broad: The truncates were the cause of the race condition I was mentioning. Postgres truncates on tables with very few rows can be very slow. They were actually causing deadlocks when there were many calls to the truncate function happening in quick succession.

internal/builds/builds_test.go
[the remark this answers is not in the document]
jyang-broad: Sounds reasonable to me, I'll look into doing that too.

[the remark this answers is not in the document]
mflinn-broad: I'm fine with just defaulting to assert though for the sake of simplicity

internal/testutils/integrationHelpers.go
jyang-broad: :mindblown: nice find!

jyang-broad: er also, I'm cool w/ just nuking the old, commented truncate statement in that case.

mflinn-broad: Thanks! Yeah if we were seeding and then dropping very large tables in tests then I think truncate would be the right tool for that job.

PR #19, opened 2021-10-04
DDO-1564 Complete Necessary functionality for service instances

mflinn-broad:
This PR implements the remaining `ServiceInstance` aka association between service and environment, needed for accelerate metrics. There are two key functionalities implemented here:

1. The ability to create new service instances. This functionality performs a check to make sure that both the service and environment exist and will create them if not before actually persisting the association to the `service_instances` table. To facilitate this functionality `FindOrCreate` methods were added to both the services and environments controllers. The combination of service and environment together are required to be unique
2. The ability to look up service instances given the environment and service names, this is a critical functionality for the eventual `deploys` route handlers.

This PR does not contain any api handlers as the service instances functionality is internal only logic and it is essentially just an associations table of service and environment ids. I don't think it necessarily makes sense to expose this implementation detail in the api directly. The `deploys` api will use this functionality under the hood.

### DIFF
--- a/internal/deploys/serviceInstance.go
+++ b/internal/deploys/serviceInstance.go
@@ -5,12 +5,27 @@ package deploys
 // that is used to build the association between a build, service, and environment
 // at a specific point in time which is needed to represent a deploy
 
-import "gorm.io/gorm"
+import (
+	"fmt"
+
+	"github.com/broadinstitute/sherlock/internal/environments"
+	"github.com/broadinstitute/sherlock/internal/services"
+	"gorm.io/gorm"
+)
 
 // ServiceInstanceController is the type used to manage logic related to working with
 // ServiceInstance entities
 type ServiceInstanceController struct {
-	store serviceInstanceStore
+	store        serviceInstanceStore
+	services     *services.ServiceController
+	environments *environments.EnvironmentController
+}
+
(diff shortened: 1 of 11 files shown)


internal/deploys/serviceInstance.go
```diff
@@ -27,6 +44,30 @@ func (sic *ServiceInstanceController) ListAll() ([]ServiceInstance, error) {
...
+	environmentID, err := sic.environments.FindOrCreate(newServiceInstance.EnvironmentName)
+	if err != nil {
+		return ServiceInstance{}, err
+	}
+
+	// check if the service already exists
+	serviceID, doesExist := sic.services.DoesServiceExist(newServiceInstance.ServiceName)
+	if !doesExist {
+		return ServiceInstance{}, fmt.Errorf("service: %s does not exist", newServiceInstance.ServiceName)
+	}
```
jack-r-warren, line 60: The comment says it will do a find or create operation for the service, but it seems like if the service doesn't exist it always throws an error

jack-r-warren (approved):

internal/deploys/serviceInstance.go
mflinn-broad: oh whoops forgot to update the that block 🤦🏻. Thanks for the catch

PR #20, opened 2021-10-05
DDO-1564 Create New Deploys and List Deploy History for Service/Environment Control Plane Functionality

mflinn-broad:
Ok, so there is a bit of chonk here as implementing the core functionality for this ticket exposed some issues in our integration test suite. 

This PR essentially does 2 things.
1. It implements the core control plane functionality to track deploy entities. These are the essential component to calculating accelerate metrics. They are essentially an association between a service, environment, and build version at a point in time. It is this functionality that will be instrumented to track metrics like deploy frequency and lead time to prod.
2. The deploy tracking functionality involves some more heavyweight database operations. This exposed some previously unnoticed database race conditions in our integration test suite. This pr also fixes those data base race conditions. It accomplishes this by running each test in its own isolated transaction that gets rolled back after the test completes. Also any logic in tests that previously expected an empty database has been removed, this more closely mirrors an actual database anyways. I nice side effect of this improvement is that the tests can all be run in parallel now without stepping on each other.

### DIFF
--- a/internal/builds/builds_test.go
+++ b/internal/builds/builds_test.go
@@ -36,14 +36,12 @@ func (suite *BuildsIntegrationTestSuite) SetupTest() {
 	suite.app = initTestApp(suite.T())
 }
 
-func (suite *BuildsIntegrationTestSuite) TearDownSuite() {
-	// ensure we clean the db at the very end
-	testutils.Cleanup(suite.T(), suite.app.db)
+func (suite *BuildsIntegrationTestSuite) TearDownTest() {
+	suite.app.db.Rollback()
 }
 
 func (suite *BuildsIntegrationTestSuite) TestCreateBuild() {
 	suite.Run("creates a new build", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		newBuild := suite.goodCreateBuildRequest
 
@@ -54,7 +52,6 @@ func (suite *BuildsIntegrationTestSuite) TestCreateBuild() {
 	})
 
 	suite.Run("fails with empty create request", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		newBuild := CreateBuildRequest{}
 
@@ -64,21 +61,15 @@ func (suite *BuildsIntegrationTestSuite) TestCreateBuild() {
 	})
 
 	suite.Run("fails on non-unique version string", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		// create a valid build
 		_, err := suite.app.builds.CreateNew(suite.goodCreateBuildRequest)
-		suite.Assert().NoError(err)
-
-		// try to create another build with the same version string
-		_, err = suite.app.builds.CreateNew(suite.goodCreateBuildRequest)
-		suite.Assert().Error(err)
+		suite.Assert().ErrorIs(err, ErrDuplicateVersionString)
 	})
 }
 
 func (suite *BuildsIntegrationTestSuite) TestGetByID() {
 	suite.Run("fails with non-existent id", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		_, err := suite.app.builds.GetByID(23)
 		suite.Require().ErrorIs(err, ErrBuildNotFound)
@@ -86,7 +77,6 @@ func (suite *BuildsIntegrationTestSuite) TestGetByID() {
 	})
 
 	suite.Run("retrives an existing build", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		newBuild := suite.goodCreateBuildRequest
 		build, err := suite.app.builds.CreateNew(newBuild)
@@ -102,7 +92,6 @@ func (suite *BuildsIntegrationTestSuite) TestGetByID() {
 
 func (suite *BuildsIntegrationTestSuite) TestGetByVersionString() {
 	suite.Run("successful looks up existing build by version string", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		// create a build instance to look up
 		existingBuild, err := suite.app.builds.CreateNew(suite.goodCreateBuildRequest)
@@ -116,7 +105,6 @@ func (suite *BuildsIntegrationTestSuite) TestGetByVersionString() {
 	})
 
 	suite.Run("errors not found for non-existent version string", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		_, err := suite.app.builds.GetByVersionString("does-not-exist")
 		suite.Assert().ErrorIs(err, ErrBuildNotFound)
@@ -132,6 +120,6 @@ func initTestApp(t *testing.T) *testApplication {
 	dbConn := testutils.ConnectAndMigrate(t)
 	return &testApplication{
 		builds: NewController(dbConn),
-		db:     dbConn,
+		db:     dbConn.Begin(),
 	}
 }
--- a/internal/builds/builds_test.go
+++ b/internal/builds/builds_test.go
@@ -25,7 +25,6 @@ func TestBuildsIntegrationSuite(t *testing.T) {
 }
 
 func (suite *BuildsIntegrationTestSuite) SetupTest() {
-	// start a new db transaction for each test
 	suite.goodCreateBuildRequest = CreateBuildRequest{
 		VersionString: faker.URL(),
 		CommitSha:     faker.UUIDDigit(),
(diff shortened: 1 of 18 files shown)


internal/builds/builds_test.go
```diff
@@ -96,13 +92,35 @@ func (suite *BuildsIntegrationTestSuite) TestGetByID() {
...
+	suite.Run("successful looks up existing build by version string", func() {
+
+		// create a build instance to look up
+		existingBuild, err := suite.app.builds.CreateNew(suite.goodCreateBuildRequest)
+		suite.Require().NoError(err)
+
+		result, err := suite.app.builds.GetByVersionString(suite.goodCreateBuildRequest.VersionString)
+		suite.Assert().NoError(err)
+
+		// make sure the id's match
```
jyang-broad, line 105: sp, s/id's/ids/

internal/builds/builds_test.go
```diff
@@ -96,13 +92,35 @@ func (suite *BuildsIntegrationTestSuite) TestGetByID() {
...
+}
+
 type testApplication struct {
 	builds *BuildController
 	db     *gorm.DB
 }
 
 func initTestApp(t *testing.T) *testApplication {
 	dbConn := testutils.ConnectAndMigrate(t)
+	dbConn = dbConn.Begin(&sql.TxOptions{Isolation: sql.LevelSerializable})
```
jyang-broad, line 123: I never ended up pushing sql-shenanigans in my last commit, but I was reading around and I think pgx implements anything database/sql and since that's our driver anyway we should prefer that. 

Also, is this movable to testutils like ConnectAndMigrate? It's a little bit scary-sql-thing that probably would scare people.

internal/deploys/deploys_test.go
```diff
@@ -0,0 +1,187 @@
...
+func (suite *DeployIntegrationTestSuite) TearDownTest() {
+	// make sure to clean db at the end
+	suite.app.db.Rollback()
+}
+
+func (suite *DeployIntegrationTestSuite) TestCreateDeploy() {
+	suite.Run("creates deploy from pre-existing service instance and build", func() {
+		// populate a build to deploy
+		existingBuildReq := builds.CreateBuildRequest{
+			VersionString: faker.URL(),
```
jyang-broad, line 54: 👍 I really should start following this pattern.

internal/deploys/models.go
```diff
@@ -51,29 +52,13 @@ func (db dataStore) createNew(environmentID, serviceID int) (ServiceInstance, er
...
-	err := db.Preload("Service").
-		Preload("Environment").
-		First(&newServiceInstance, newServiceInstance.ID).Error
-
-	return newServiceInstance, err
+	return newServiceInstance, nil
 }
 
-func (db dataStore) getByEnvironmentAndServiceName(environmentName, serviceName string) (ServiceInstance, error) {
+func (db dataStore) getByEnvironmentAndServiceID(environmentID, serviceID int) (ServiceInstance, error) {
```
jyang-broad, line 58: what's your thoughts on requiring comments on all methods. I noticed we do it pretty consistently in Controllers, but not in our models, I'm of the opinion of removing the decision of whether to have a comment in favor of always having one (at the understandable cost of more chatty code).

internal/deploys/models.go
```diff
@@ -87,3 +72,51 @@ func (db dataStore) listAll() ([]ServiceInstance, error) {
...
+}
+
+func (db dataStore) getDeploysByServiceInstance(serviceInstanceID int) ([]Deploy, error) {
+	var deploys []Deploy
+
+	err := db.Preload("ServiceInstance").
+		Preload("ServiceInstance.Service").
+		Preload("ServiceInstance.Environment").
+		Preload("Build").
+		Preload("Build.Service").
```
jyang-broad, line 117: I haven't really looked too much into the lazy-loading behavior of GORM but maybe here is a good time to do that, this just feels massive, also we're loading service twice. Or least (considering time) maybe put a TODO comment to note we reaaaaaally should make sure this is won't explode us.

internal/deploys/serviceInstance_test.go
```diff
@@ -120,29 +112,17 @@ func (suite *ServiceInstanceIntegrationTestSuite) TestCreateServiceInstance() {
 		result, err := suite.app.serviceInstances.CreateNew(newServiceInstanceReq)
 		suite.Require().NoError(err)
 
-		suite.Assert().Equal(newServiceInstanceReq.ServiceName, result.Service.Name)
+		suite.Assert().NotEqual(0, result.ServiceID)
```
jyang-broad, line 115: nothing wrong here, just me expressing my annoyance that golang forces us to do this. I'm very thankful that no SQL databases start with ID 0.

internal/environments/environments_test.go
```diff
@@ -181,45 +187,44 @@ func (suite *EnvironmentTestSuite) TestIntegrationEnvironmentGetByName() {
 
 func (suite *EnvironmentTestSuite) TestIntegrationEnvironmentListAll() {
 	suite.Run("ListAll returns nothing", func() {
-		testutils.Cleanup(suite.T(), suite.testApp.db)
 
 		foundEnvironments, err := suite.testApp.Environments.ListAll()
 
-		assert.Equal(suite.T(), len(foundEnvironments), 0)
+		assert.GreaterOrEqual(suite.T(), len(foundEnvironments), 0)
```
jyang-broad, line 193: wouldn't this pass incorrectly if ListAll returned 1 or more?

internal/environments/environments_test.go
```diff
@@ -181,45 +187,44 @@ func (suite *EnvironmentTestSuite) TestIntegrationEnvironmentGetByName() {
...
+
+		_, err = suite.testApp.Environments.CreateNew(randomEnvRequest)
 		assert.NoError(suite.T(), err)
 		_, err = suite.testApp.Environments.CreateNew(suite.anotherEnvironmentRequest)
 		assert.NoError(suite.T(), err)
 
 		foundEnvironments, err := suite.testApp.Environments.ListAll()
 
-		assert.Equal(suite.T(), len(foundEnvironments), 2)
+		assert.GreaterOrEqual(suite.T(), len(foundEnvironments), 3)
```
jyang-broad, line 221: I think I'm missing something, why are we expecting 3+ environments?

internal/services/services_test.go
```diff
@@ -59,16 +61,14 @@ func (suite *ServicesIntegrationTestSuite) TestCreateService() {
 func (suite *ServicesIntegrationTestSuite) TestListServices() {
 	// make some create service requests and populated them with fake data
 	suite.Run("ListAll returns nothing", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
 
 		services, err := suite.app.services.ListAll()
 
-		suite.Assert().Equal(len(services), 0)
+		suite.Assert().GreaterOrEqual(len(services), 0)
```
jyang-broad, line 67: this always passes, right?

jyang-broad (approved):
Looks pretty good! All the new code looks solid, but I have some concerns w/ some of the tests and the usage of GreaterOrEqual, I think a lot of them will pass incorrectly and there's potential of not being able to catch wierd unexpected corner cases where maybe something will create duplicates (or something) further down the line.

internal/builds/builds_test.go
```diff
@@ -48,41 +51,34 @@ func (suite *BuildsIntegrationTestSuite) TestCreateBuild() {
...
 
-		_, err := suite.app.builds.CreateNew(newBuild)
+	_, err := suite.app.builds.CreateNew(newBuild)
 
-		suite.Require().Error(err)
-	})
-
-	suite.Run("fails on non-unique version string", func() {
-		testutils.Cleanup(suite.T(), suite.app.db)
+	suite.Require().Error(err)
```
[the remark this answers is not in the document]
mflinn-broad: nope:)

internal/builds/builds_test.go
```diff
@@ -22,24 +24,25 @@ func TestBuildsIntegrationSuite(t *testing.T) {
 	suite.Run(t, new(BuildsIntegrationTestSuite))
 }
 
-func (suite *BuildsIntegrationTestSuite) SetupSuite() {
-	suite.app = initTestApp(suite.T())
-	// ensure the db is clean before running suite
-	testutils.Cleanup(suite.T(), suite.app.db)
+func (suite *BuildsIntegrationTestSuite) SetupTest() {
+	// start a new db transaction for each test
```
jack-r-warren, line 28: Maybe this comment should reference that this is an effect of initTestApp specifically--it wasn't until I looked there that I understood what was going on here, I thought this comment was misplaced

jack-r-warren (approved):

internal/environments/environments_test.go
jyang-broad: oh sorry, I was meaning some wierd unexpected corner case where we end up with some strange number of new entries (not specifically checking for dupes, agree w/ your assessment there on mixing concerns), I just couldn't come up with a good example becuase.... wierd unexpected corner cases

choover-broad: > That said there are some assumptions we had in the list all tests previously that are no longer valid. The main one being that the db is empty. It could be completely empty but it might not.

For my own edification: why is this the case if all db changes for each test happen in a transaction that is rolled back?

PR #22, opened 2021-10-12
Ddo 1565 Add Deploy Handlers and Metrics Support

mflinn-broad:


internal/deploys/deploys_test.go
```diff
@@ -120,7 +120,7 @@ func (suite *DeployIntegrationTestSuite) TestCreateDeploy() {
 		}
 
 		_, err := suite.app.deploys.CreateNew(newDeployReq)
-		suite.Assert().ErrorIs(err, builds.ErrBuildNotFound)
+		suite.Assert().NoError(err)
```
jyang-broad, line 123: I think we need an assert to check the build exists as the  test desc states.

internal/deploys/handler_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/broadinstitute/sherlock/internal/builds"
+	"github.com/broadinstitute/sherlock/internal/testutils"
+	"github.com/bxcodec/faker/v3"
+)
+
+func (suite *DeployIntegrationTestSuite) TestCreateDeployHandler() {
+	// setup prepopulate a service instance and some builds
+	var preExistingServiceInstanceReq CreateServiceInstanceRequest
+	err := faker.FakeData(&preExistingServiceInstanceReq)
+	suite.Require().NoError(err)
```
jyang-broad, line 18: conflicted here on consistency vs conciseness, what's your thoughts on 
```suite.Require().NoError(faker.FakeData(&preExistingServiceInstanceReq))```
considering we can't do that for dual output methods.

File: internal/deploys/handler_test.go [hunk 3]
```diff
@@ -0,0 +1,138 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/broadinstitute/sherlock/internal/builds"
+	"github.com/broadinstitute/sherlock/internal/testutils"
+	"github.com/bxcodec/faker/v3"
+)
+
+func (suite *DeployIntegrationTestSuite) TestCreateDeployHandler() {
+	// setup prepopulate a service instance and some builds
```
jyang-broad, line 15: setup prepopulate <- redundant words.

internal/deploys/handler_test.go
```diff
@@ -0,0 +1,138 @@
...
+		reqBody := new(bytes.Buffer)
+		err := json.NewEncoder(reqBody).Encode(deployReq)
+		suite.Require().NoError(err)
+		path := fmt.Sprintf("/deploys/%s/%s", preExistingServiceInstance.Environment.Name, preExistingServiceInstance.Service.Name)
+		req, err := http.NewRequest(http.MethodPost, path, reqBody)
+		suite.Require().NoError(err)
+		ctx.Request = req
+
+		suite.app.deploys.createDeploy(ctx)
+		suite.Assert().Equal(http.StatusCreated, response.Code)
```
jyang-broad, line 52: I think we return the serialized deploy here so we should confirm it's correctly updated and not half-formed somehow.

internal/deploys/handlers.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (dc *DeployController) getDeploysByEnvironmentAndService(c *gin.Context) {
+	environment := c.Param("environment")
+	service := c.Param("service")
+
+	deploys, err := dc.GetDeploysByEnvironmentAndService(environment, service)
+	if err != nil {
+		switch err {
+		case ErrServiceInstanceNotFound:
+			c.JSON(http.StatusNotFound, Response{Error: err.Error()})
```
jyang-broad, line 29: we should probably at least check that 404 works, the 500 is more :shrug:

internal/deploys/serializers.go
```diff
@@ -49,3 +58,44 @@ func (sis *ServiceInstancesSerializer) Response() []ServiceInstanceResponse {
 	}
 	return serviceInstances
 }
+
+// DeployResponse is the type used for generating api responses
+// containging information about deploy(s)
```
jyang-broad, line 63: s/containging/containing

PR #28, opened 2021-11-01
DDO-1684 Fix for Artificially increasing lead times

mflinn-broad:
Previously sherlock was increasing lead time artificially when the same build was deployed to an environment multiple times in a row by recalculating the time diff between image creation and current deploy time for each redeploy. This PR adds a check so that lead time will only be updated if a new build version is being deployed. These re-deploy events are still recorded in sherlock's db and count towards deploy frequency.

### DIFF
--- a/internal/deploys/deploys.go
+++ b/internal/deploys/deploys.go
@@ -79,6 +79,16 @@ func (dc *DeployController) GetDeploysByEnvironmentAndService(environmentName, s
 	return dc.store.getDeploysByServiceInstance(serviceInstance.ID)
 }
 
+// GetMostRecentDeploy will look up the most recent ie currently active deploy for a given service instance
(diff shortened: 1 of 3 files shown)


jack-r-warren (approved):

choover-broad (approved):
LGTM!

One thought: will rollbacks still trigger this issue?

eg. build 50 goes out, we roll back to build 49 because of a bug, then roll forward to 51...

mflinn-broad:
> LGTM!
> 
> One thought: will rollbacks still trigger this issue?
> 
> eg. build 50 goes out, we roll back to build 49 because of a bug, then roll forward to 51...

Great point! I think I'll address this case separately in a different. As this scenario might require a little more thought  / logic to account for.

jyang-broad (approved):
Look sane! Needs tests though.

PR #30, opened 2021-11-22
DDO-1691  Reorganize code into M(V?)C packages

mflinn-broad:
This is a big looking PR that is essentially a no-op. There is no changed or added functionality here. 

All that is happening is that sherlock's database modeling code is being moved to a common `package models` rather than being distributed across multiple functional packages. 

This is important for planned future features for sherlock that will require the data models to reference eachother. This enables them to do so without causing import cycles.


choover-broad (approved):
this looks nice! ✔️ to unblock, will try to loop back for a closer look later!

PR #33, opened 2022-04-26
[DDO-2041] Biiig refactor to complete the "M(V?)C" move that was started a while back

jack-r-warren:
A while back, the models were centralized into a `models` package, and the controller for just the Cluster type was moved into a `controllers` package--Cluster lacks a direct API, so it is the simplest case.

This PR untangles the controllers, handlers, and serializers for all the other datatypes to fit them into the MVC-ish pattern (the "-ish" is because we don't quite have a view here, unless you consider API handlers the view, in which case we do have that). It also organizes them into "v1" subpackages to avoid naming conflicts with the stuff from the [doc](https://docs.google.com/document/d/1TINiZTzc6Xfo4T0FEjBGaqbqe68wMmhVL8fj_xs1m08/edit).

There's no added or removed functionality here, all tests pass just as before. I'll highlight the main changes:

<details>
<summary>
Directory structure
</summary>

Folders for `builds`, `deploys`, `environments`, and `services` are gone, and their constituent files have been spread out into the existing `controllers` and the new `handlers` and `serializers`. Mocks went into `controllers` based on their imports (couldn't go anywhere else) and the unused-but-good-for-examples seed functions went into `models`.

```
.
├── README.md
├── cli
│   ├── builds.go
│   ├── create_build.go
│   ├── create_build_test.go
│   ├── create_deploy.go
│   ├── create_deploy_test.go
│   ├── deploys.go
│   ├── root.go
│   ├── version.go
│   └── version_test.go
├── controllers
│   └── v1controllers
│       ├── allocation_pools.go
│       ├── allocation_pools_test.go
│       ├── builds.go
│       ├── builds_test.go
│       ├── cluster.go
│       ├── clusters_test.go
│       ├── deploys.go
│       ├── deploys_test.go
│       ├── environments.go
│       ├── environments_mock.go
│       ├── environments_test.go
│       ├── service_instance.go
│       ├── service_instance_mock.go
│       ├── service_instance_test.go
│       ├── services.go
│       ├── services_mock.go
│       ├── services_test.go
│       └── test_helpers.go
├── db
│   ├── README.md
│   └── db.go
├── handlers
│   └── v1handlers
│       ├── builds_handlers.go
│       ├── builds_handlers_test.go
│       ├── deploys_handlers.go
│       ├── deploys_handlers_test.go
│       ├── environments_handlers.go
│       ├── environments_handlers_test.go
│       ├── services_handlers.go
│       └── services_handlers_test.go
├── metrics
│   └── metrics.go
├── models
│   └── v1models
│       ├── allocation_pools.go
│       ├── builds.go
│       ├── builds_seed.go
│       ├── cluster.go
│       ├── deploys.go
│       ├── deploys_seed.go
│       ├── environments.go
│       ├── environments_seed.go
│       ├── service_instances.go
│       ├── services.go
│       └── services_seed.go
├── serializers
│   └── v1serializers
│       ├── builds_serializers.go
│       ├── deploys_serializers.go
│       ├── environments_serializers.go
│       └── services_serializers.go
├── sherlock
│   ├── routes.go
│   └── sherlock.go
├── testutils
│   ├── helpers.go
│   └── integrationHelpers.go
└── version
    └── version.go
```

</details>

<details>
<summary>
Handlers are no longer object-oriented
</summary>

In Go, methods must be defined in the same package the receiving type is defined in. This means that if handlers are written as methods on controllers, they and their tests must be in the same package as the controllers.

> Grumble grumble if you have side-effect imports like Go does you don't need this restraint for type safety, it appears to just be for "simplicity"

So, while a handler before was like this...

```go
func (dc *DeployController) getDeploysByEnvironmentAndService(c *gin.Context) {
    // ...
	c.JSON(http.StatusOK, Response{Deploys: dc.Serialize(deploys...)})
}
```

... it now has partial application like this ...

```go
func getDeploysByEnvironmentAndService(dc *v1controllers.DeployController) func(c *gin.Context) {
	return func(c *gin.Context) {
       // ...
		c.JSON(http.StatusOK, v1serializers.DeploysResponse{Deploys: dc.Serialize(deploys...)})
	}
}
```

This is because handlers must be exactly a `func(c *gin.Context)`. Before, our methods could masquerade as a handler by having a receiver of the controller, but now, we need a function that returns a handler. IMO, fair price to pay for being able to have all handlers in a separate package. There's some minor downstream impacts of this--as in, a test that used to say `dc.getDeploysByEnvironmentAndService(ctx)` would now say `getDeploysByEnvironmentAndService(dc)(ctx)`.

</details>

<details>
<summary>
Go bumped to 1.18
</summary>

This affected the Docker files. The old ones used alpine 3.14 and Go 1.16, but to go to Go 1.18 required a bump to alpine 3.15.

Go 1.18 also tries to be slick and make automatic use of the git executable if you build from a git repo (a worse version of the ldflags version injection we do, because it's just commit hash), so we have to turn that off unless we want to install and ship git in our images.

</details>

<details>
<summary>
Tweak to make migrations from test files not dependent on the test file's locations
</summary>

The existing mechanism ended up relying on relative paths from the test file to the changelog file, but with the relative path hardcoded inside the centralized test helpers. Rather than parameterizing it per test, I change the test helpers a tiny bit to effectively center the relative path on itself, rather than the calling test file.

```go
// ProjectRootFilePath is a Golang-native mechanism to get the root of the package structure
// on the filesystem. With this, tests can introspect files (e.g. database changelogs) without
// needing to use relative paths themselves (which would be relative to the test file itself).
// H/T https://stackoverflow.com/a/58294680
var (
	_, b, _, _          = runtime.Caller(0)
	ProjectRootFilePath = filepath.Join(filepath.Dir(b), "../..")
)
```

```go
	if err := db.ApplyMigrations(filepath.Join(ProjectRootFilePath, "db/migrations"), config); err == migrate.ErrNoChange {
```

</details>


internal/controllers/v1controllers/allocation_pools.go
```diff
@@ -1,27 +1,21 @@
-// AllocationPoolController, public interface for AllocationPool objects
-
-// Package allocationPools defines data structure representing
-// a allocationPool instance and methods for interacting with them
-// it is left to concrete implementations in package db or others to
-// implement these interfaces
-package allocationpools
+package v1controllers
```
jyang-broad, line 1: update comment? I'm pushing to have a top-level comment to describe what goes (or shouldn't go) in a file.

jyang-broad (approved):
Looks Sane!

PR #35, opened 2022-07-21
Sherlock Dev :chonk:

jack-r-warren:
This PR implements the "Data Types" section of the ["Putting terra-helmfile Into Sherlock, API Edition" document](https://docs.google.com/document/d/1TINiZTzc6Xfo4T0FEjBGaqbqe68wMmhVL8fj_xs1m08/edit#heading=h.oh92uxequv5y).

It doesn't yet implement the "Procedures" section of that document, nor does it do metrics generation from the v2 side yet. The CLI doesn't yet know how to talk to v2.

<h1>Changes:</h1>

<h3>

```diff
+ internal/models/v2models
```

</h3>

New types added for Cluster, Environment, Chart, Chart Version, App Version, Chart Release, and Chart Deploy Record. The tables look like this:

Screen Shot 2022-08-10 at 12 16 36 PM

> **Design changes compared to v1**
> 
> The v2 model performs data validation above the SQL level, including cases like:
> - Empty or invalid fields
> - Duplicate "selectors" (more information [here](https://docs.google.com/document/d/1TINiZTzc6Xfo4T0FEjBGaqbqe68wMmhVL8fj_xs1m08/edit#heading=h.dc5esgdqldli); summary is that this prevents duplicate names, duplicate charts in the same namespace, etc.)
> - Insufficient user access for write operations
>
> There's soft deletion here too, courtesy of Gorm. The downside is that deleting and recreating something with the same name will require us to intervene to resolve the SQL validation issue.

> **There's generics here!**
> 
> `model_store.go` defines a generic set of verbs--create, get, edit, etc.--and the other files for the actual types just implement the underlying nuts and bolts they need to. `store_set.go` is everything packaged together, so it can be passed around in tests and the controller.

<h3>

```diff
+ internal/controllers/v2controllers
```

</h3>

These controllers define the user types that are actually visible through the API. They essentially map verbs the model provides (in terms of the model types) to the same verbs defined in terms of the more friendly user types.

> **Design changes compared to v1**
> 
> The v1 controllers are defined in terms of the model types and have user type serialization elsewhere, while the new v2 controllers are defined in terms of user types themselves. This is because going from user types to model types actually requires access to the database, to resolve associations.
>
> The v2 controllers also handle defaults for incoming data.

> **There's generics here!**
> 
> `model_controller.go` defines a generic set of verbs based on those from the model, and the other files provide the user-visible types and implement the underlying nuts and bolts they need to. `controller_set.go` is everything packaged together, so it can be passed around in tests.

<h3>

```diff
+ internal/handlers/v2handlers
```

</h3>

These handlers choose what verbs from the controllers to expose via the API, and they provide documentation picked up for the Swagger page.

[You can check out the handlers and overall app behavior by visiting Sherlock Dev.](https://sherlock-dev.dsp-devops.broadinstitute.org/)

> **There's generics here!**
> 
> `generic_handlers.go` is a set of helper functions that can each go from Gin to a particular verb of a given controller. Other files expose the actual documented handlers but call the helpers to do the actual work.

<h2>Smaller Stuff:</h2>

<h3>

```diff
+ internal/auth
```

</h3>

This package defines the concept of a user contacting Sherlock. It has middleware that creates User objects from IAP headers and a rotating cache of Firecloud accounts. The User objects are passed through the rest of the program by the handlers, and they expose methods to enforce suitability when necessary.

For service accounts, emails can be provided in Sherlock's config file to grant them "suitability" without them actually having access to the Firecloud org.

<h3>

```diff
+ internal/errors
```

</h3>

This package provides some string constants that can be included in error messages generated anywhere in Sherlock. If these errors are passed back to the user, those string constants get detected and the response code of the error gets set automatically.

This means that the model itself, for example, can control whether an error it returns means a 404, 400, 403, 500, etc.

<h3>

```diff
+ internal/config
```

</h3>

Sherlock now uses Koanf for configuration and Zerolog for logging. Default configuration is embedded inside Sherlock's binary (in the top-level `config`), there's an additional layer of test configuration that is pulled in automatically (this means Sherlock tests can now natively be run from GoLang), and logging switches between CLI-style and JSON.

<h3>

```diff
~ internal/db
```

</h3>

This package provides database connection logic that ties in config, migration files, optional auto-migration from Gorm for iteration during development, and database truncation for use in tests. It's been rewritten so that we don't need to have a separate version of it elsewhere for use in tests.

It also has some retry logic so Sherlock can start before its database (no need for sleeps in the server Dockerfile).

<h3>

```diff
~ db
```

</h3>

Database migration files are now embedded inside Sherlock's binary, so this directory is now a Go package. This helps run migrations from tests from GoLand without worrying about relative paths.

<h3>

```diff
+ cmd/server/liveness.go
```

</h3>

Since I embedded the database migration files into Sherlock's binary, we won't be able to run preflight migrations. I added a basic liveness endpoint on port 8081 instead, so we can use that to avoid killed migrations. The liveness check comes online once a database connection is established but before any migrations start running.

<h3>

```diff
+ docs
```

</h3>

This files in this package are auto-generated by Swaggo--it parses comments in the handlers and structs from the rest of the program to generate OpenAPI documentation files.

I haven't automated this yet; you can use `go install github.com/swaggo/swag/cmd/swag@latest` to install Swaggo and `make generate-swagger` to regenerate the files.

<h3>

```diff
~ build
```

</h3>

I tweaked both Dockerfiles, they skip running tests since we do that elsewhere and they cache dependencies so building locally is about 40 seconds faster.


mflinn-broad (approved):
I'm not going to go line by line through every code change, but from a high level architectural standpoint everything looks good to me. I may follow up in this PR with a few more questions but approving to unblock.

build/cli/Dockerfile
```diff
@@ -1,16 +1,21 @@
-ARG GO_VERSION='1.18'
+ARG GO_VERSION='1.19'
 ARG ALPINE_VERSION='3.15'
```
mflinn-broad, line 2: Latest alpine version is 3.16, might as well bump while we are updating versions?

internal/controllers/v2controllers/controller_set.go
```diff
@@ -0,0 +1,25 @@
+package v2controllers
+
+import "github.com/broadinstitute/sherlock/internal/models/v2models"
```
mflinn-broad, line 3: Unused import?

mflinn-broad (approved):

internal/controllers/v2controllers/controller_set.go
jack-r-warren: I don't think so, the ControllerSet constructor accepts a v2models.StoreSet that it passes to the internal controller constructors

PR #59, opened 2022-08-25
[DDO-2295] Cluster name presence accidentally enforced by Swagger

jack-r-warren:
This is enforced (there's a test for it) internally, and at the Swagger level it's actually bad because it prevents us from listing all clusters. At one point I'd tried to remove all the validate struct tags and must have just missed one.


mflinn-broad (commented):
retroactive thumbs up

PR #68, opened 2022-09-21
[DDO-2346] Error faster on Changeset Chart mismatches

jack-r-warren:
This would error out elsewhere but we can fail-fast and give a good error message by catching this case.

I’ve actually had this code locally forever, just forgot to push it.


jyang-broad (approved):
Looks sane! Nice error messages.
Adds 2 new explicit error cases, and some tests. codecov complaining though.

PR #75, opened 2022-10-11
DDO-2400 New Fields to support Thelma State

mflinn-broad:
This PR adds support for a few different fields on various sherlock resources.

1. `cluster.location` this is the cloud region and or zone that a cluster is located in. Thelma needs this info in order to set kubectx properly.
2. `environment.helmfileRef` and `cluster.helmfileRef`. Both of these are needed for argocd manifest rendering in Thelma
3. `firecloudDevelopRef` on both chart-releases and changesets. This is also required for argocd manifest for firecloud-devleop config rendering.

Looking for feedback on:

1. Did I understand V2 sherlock correctly
2. Am I using the `Creatable vs Editable` generic types in the right way.
3. I want to make sure I'm not making a breaking change to changesets.

I'm pretty sure I set defaults for all these so these changes shouldn't be breaking for anything existing 🤞🏻 

Open design question: Wouid it make sense to have a `DefaultFirecloudDevelopRef` on `Environment` which could be used to set a dynamic default to the correct firecloud-develop branch for live envs?


internal/models/v2models/chart_release_version.go
```diff
@@ -214,6 +216,10 @@ func (chartReleaseVersion *ChartReleaseVersion) validate() error {
 		return fmt.Errorf("a %T must have a terra-helmfile git reference", chartReleaseVersion)
 	}
 
+	if chartReleaseVersion.FirecloudDevelopRef == nil || *chartReleaseVersion.FirecloudDevelopRef == "" {
+		return fmt.Errorf("a %T must have a firecloud-develop git reference", chartReleaseVersion)
+	}
```
jack-r-warren, line 221: I'm not sure this quite makes sense--every single chart release _must_ have a firecloud-develop ref?

internal/models/v2models/cluster.go
```diff
@@ -101,6 +104,9 @@ func validateCluster(cluster *Cluster) error {
 	if cluster.Address == nil || *cluster.Address == "" {
 		return fmt.Errorf("a %T must have a non-empty address", cluster)
 	}
+	if cluster.Location == "" {
+		return fmt.Errorf("a %T must specify a location", cluster)
+	}
```
jack-r-warren, line 109: Should we validate that helmfile ref isn't empty?

db/migrations/000020_add_helmfile_and_fc_develop_ref_fields.up.sql
```diff
@@ -0,0 +1,17 @@
...
+    add if not exists to_firecloud_develop_ref text;
+
+alter table v2_clusters
+    add if not exists helmfile_ref text;
+
+alter table v2_environments
+    add if not exists helmfile_ref text;
+
+alter table v2_clusters
+    add if not exists location text;
```
jack-r-warren, line 17: Gorm and the migration files are inconsistent here. The gorm structs say that some of these columns need not null constraints and defaults, but the migration files don't uphold that

internal/controllers/v2controllers/changeset.go
```diff
@@ -277,6 +284,8 @@ func setChangesetDynamicDefaults(changeset *Changeset, stores *v2models.StoreSet
 		changeset.FromChartVersionReference = strconv.FormatUint(uint64(*chartRelease.ChartVersionID), 10)
 	}
 	changeset.FromHelmfileRef = chartRelease.HelmfileRef
+	// changesets should not modify firecloud-develop refs
```
jack-r-warren, line 287: The code right here is fine, but I think we should think through this comment.

The way I've been thinking of it, the only place where we want to treat firecloud-develop refs differently from terra-helmfile is line 57 and 136 of v2controllers/changeset_procedures.go. Literally everywhere else should treat them the same (with the exception of maybe not making it a required field), and that will give the behavior of "you can change them when you ask to change them specifically, but they don't automatically get changed when you do any sort of promotion".

Right now, the field can be set upon creation, but it isn't fully looped through the changeset process enough--we can't mutate it normally.

db/migrations/000020_add_helmfile_and_fc_develop_ref_fields.up.sql
mflinn-broad: Ah good catch, I added the gorm stuff as I went, and forgot to go back and update this

internal/controllers/v2controllers/changeset.go
mflinn-broad: Got it I misunderstood what you meant when you said "leave it out of the changeset stuff" yesterday then. I kind of had to take a guess at what that meant. I can update so it's more similar to everything else.

internal/models/v2models/environment.go
```diff
@@ -25,6 +26,7 @@ type Environment struct {
 	RequiresSuitability *bool
 	BaseDomain          *string
 	NamePrefixesDomain  *bool
+	HelmfileRef         *string `gorm:"not null; default:master"`
```
jack-r-warren, line 29: Same comment as for cluster -- the default here is inconsistent with the controller and we should check that the field doesn't get set to empty

internal/controllers/v2controllers/changeset.go
jack-r-warren: Yeah I think I wasn't specific enough, that's on me. Those specific two places are where the "use exact versions from" stuff is handled and if we omit it there I think it'll do what we want

internal/models/v2models/cluster.go
```diff
@@ -13,10 +14,12 @@ type Cluster struct {
 	Provider          string `gorm:"not null; default:null"`
 	GoogleProject     string
 	AzureSubscription string
+	Location          string
 	// Mutable
 	Base                *string `gorm:"not null; default:null"`
 	Address             *string `gorm:"not null; default:null"`
 	RequiresSuitability *bool   `gorm:"not null; default:null"`
+	HelmfileRef         *string `gorm:"not null; default:master"`
```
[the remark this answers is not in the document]
mflinn-broad: I removed the defaults at this level, it's not needed for migration, this was a left over from some of my earliest commits on this PR and I forgot about it.

internal/controllers/v2controllers/changeset.go
mflinn-broad: So are you saying to just not include a similar line to 57 and and 136 at all for firecloud-develop ref?

db/migrations/000020_add_helmfile_and_fc_develop_ref_fields.up.sql
```diff
@@ -0,0 +1,17 @@
...
+    add if not exists firecloud_develop_ref text;
+
+alter table v2_changesets
+    add if not exists from_firecloud_develop_ref text;
+
+alter table v2_changesets
+    add if not exists to_firecloud_develop_ref text;
+
+alter table v2_clusters
+    add if not exists helmfile_ref text NOT NULL;
```
jack-r-warren, line 11: without a default here, this will fail on database rows that already exist before the migration happens

internal/models/v2models/chart_release_version_test.go
```diff
@@ -97,7 +100,8 @@ func TestChartReleaseVersion_validate(t *testing.T) {
 				},
 				ChartVersionID: testutils.PointerTo[uint](2),
 
-				HelmfileRef: testutils.PointerTo("e5f6g7h8"),
+				HelmfileRef:         testutils.PointerTo("e5f6g7h8"),
+				FirecloudDevelopRef: testutils.PointerTo("dev"),
```
jack-r-warren, line 104: if fc-dev ref is no longer a mandatory field I think we should omit it at some points in these validation tests to cover that case

db/migrations/000020_add_helmfile_and_fc_develop_ref_fields.up.sql
jack-r-warren: i wonder if there's a way to add a column and use a default during that add operation but not leave the default configuration laying around in the sql afterwards. like we want a one time "put HEAD here in order to add the column"

mflinn-broad: hmm, I bet there is a way to do that

mflinn-broad: I've thought of plenty of hacky ways to do this but ultimately we should probably just set a default.

mflinn-broad: [this should work](https://github.com/broadinstitute/sherlock/pull/75/commits/4d75eb432737150aa33dd9f054aa5d8cc0eae2f8)

db/migrations/000020_add_helmfile_and_fc_develop_ref_fields.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    add if not exists from_firecloud_develop_ref text;
+
+alter table v2_changesets
+    add if not exists to_firecloud_develop_ref text;
+
+alter table v2_clusters
+    add if not exists helmfile_ref text DEFAULT 'HEAD' NOT NULL;
+
+alter table v2_clusters
+    alter column helmfile_ref DROP DEFAULT;
```
jack-r-warren, line 14: okay i actually really like this, didn't think of this lol

jack-r-warren (approved):

jack-r-warren (approved):
looks perfect

PR #78, opened 2022-10-18
[DDO-2414] Add a latest tag to Sherlock's builds

jack-r-warren:


jyang-broad (approved):
adds latest tag to the tagging step on the sherlock build for both CLI and server

PR #86, opened 2022-11-02
DDO-2445 Refactor Lead Time Metrics To Support Multi Replica Sherlock

mflinn-broad:
**Problem:** Right now with the way Sherlock's accelerate metrics implementation works, the metrics would report incorrectly if we run more than one Sherlock instance.

For the case of the deploy frequency metric this can be solved with just a minor adjustment to the Prometheus query. In iterator logic this would look like:
```
sherlock_instances
    .iterate()
    .map(delta(sherlock_deploy_frequency{service="blah",env="blah}))
    .sum()
```

However for lead times it is a little more complex. With the event driven lead time metric as we currently have it only the sherlock replica that serves the request will be aware of the updated lead time, the others will still have the old lead time which will cause weird looking data in grafana

**Solution:** Refactor the lead time metric implementation so that rather than responding in an event based  manner on requests to a particular endpoint. Sherlock will instead run a concurrent process which maintains a cache of the most recent lead time records in the db. The cache is flushed and rebuilt on some interval. The metrics write interval and cache flush interval are both changeable via Sherlock config.

This implementation uses [time.Tickers](https://gobyexample.com/tickers) to coordinate this.

There is potential that this could cause a bit of jitter in the lead time time series data in grafana if the tickers are out of sync in different sherlock instances. However this can be solved by a combination of playing with the timer configs and some smoothing on the prometheus side. Millisecond accuracy is not a requirement of accelerate metrics.


choover-broad (approved):
Nice!

mflinn-broad:
Thanks, for the thumb. Just realized I have a major bug here. Will fix.

PR #101, opened 2022-12-08
[DDO-2520] New client actions

jack-r-warren:
- Adds client-get-environment, which is a GHA-native mechanism to grab some info about an environment. Can easily extend it over time--just means we don't need to write curls and bash everywhere.
- Extends client-set-environment-app-version, so if you pass a GH token it'll call terra-github-workflows to sync whatever got updated. It uses client-get-environment to be smart about how it treats templates just to avoid the user gotcha.
  - client-set-environment-app-version already had parity with terra-helmfile's update-service, but now it actually has full parity with fc-jenkins's gke-update-service, which is neat.

### DIFF
--- a/.github/workflows/client-get-environment.yaml
+++ b/.github/workflows/client-get-environment.yaml
@@ -1,0 +1,90 @@
+name: Get Environment
+
+on:
+  workflow_call:
+
+    inputs:
+
+      ##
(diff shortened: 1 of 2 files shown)


mflinn-broad (approved):
cool cool

PR #112, opened 2023-01-03
DDO-2550 AutoDelete adjustments to fix zero-value client-side problems

choover-broad:
This single-line PR addresses an issue with the `--delete-after` setting not propagating from `thelma bee create` when version overrides with eg. `--terra-helmfile-ref` are also provided.

### Existing behavior
A comment on the `AutoDelete` field causes `swaggo` to produce a Swagger doc that nests the autodelete type under `allOf`:

```
            "properties": {
                "autoDelete": {
                    "description": "Used to schedule automatic deletion of BEEs",
                    "allOf": [
                        {
                            "$ref": "#/definitions/environment.AutoDelete"
                        }
                    ]
                },
```

This in turn causes go-swagger to generate a client where environment.AutoDelete is bundled into a non-nullable struct:

```
type V2controllersEnvironment struct {

	// Used to schedule automatic deletion of BEEs
	AutoDelete struct {
		EnvironmentAutoDelete
	} `json:"autoDelete,omitempty"`

```

### Fix
Removing the comment creates a document with a direct reference, like so:
```
                "autoDelete": {
                    "$ref": "#/definitions/environment.AutoDelete"
                },
```

which results in a generated client with a simple, nullable *EnvironmentAutoDelete
```
// swagger:model v2controllers.Environment
type V2controllersEnvironment struct {

        // auto delete
        AutoDelete *EnvironmentAutoDelete `json:"autoDelete,omitempty"`

```

Why does it matter that this field is nullable? Because if AutoDelete is not nullable, every PATCH request to environment includes an empty AutoDelete field like this:
```
{
  "autoDelete": {
    "after": <zero-time in UTC>,
  }
}
```

Sherlock interprets this as "`enabled` is false, so I should remove the environment's auto-delete settings". (It would be better for Sherlock to return an error indicating this payload is invalid, but after some investigating, fixing that seems like it would involve digging deep into the controller code, and the client would still need to be fixed).

### Related
* https://github.com/broadinstitute/thelma/pull/101

### DIFF
--- a/internal/controllers/v2controllers/environment.go
+++ b/internal/controllers/v2controllers/environment.go
@@ -43,7 +43,7 @@ type EditableEnvironment struct {
 	NamePrefixesDomain         *bool                   `json:"namePrefixesDomain" form:"namePrefixesDomain" default:"true"`
 	HelmfileRef                *string                 `json:"helmfileRef" form:"helmfileRef" default:"HEAD"`
 	PreventDeletion            *bool                   `json:"preventDeletion" form:"preventDeletion" default:"false"` // Used to protect specific BEEs from deletion (thelma checks this field)
-	AutoDelete                 *environment.AutoDelete `json:"autoDelete" form:"autoDelete"`                           // Used to schedule automatic deletion of BEEs
+	AutoDelete                 *environment.AutoDelete `json:"autoDelete" form:"autoDelete"`
 	Description                *string                 `json:"description" form:"description"`
 }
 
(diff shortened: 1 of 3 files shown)


internal/controllers/v2controllers/environment.go
```diff
@@ -43,7 +43,7 @@ type EditableEnvironment struct {
 	NamePrefixesDomain         *bool                   `json:"namePrefixesDomain" form:"namePrefixesDomain" default:"true"`
 	HelmfileRef                *string                 `json:"helmfileRef" form:"helmfileRef" default:"HEAD"`
 	PreventDeletion            *bool                   `json:"preventDeletion" form:"preventDeletion" default:"false"` // Used to protect specific BEEs from deletion (thelma checks this field)
-	AutoDelete                 *environment.AutoDelete `json:"autoDelete" form:"autoDelete"`                           // Used to schedule automatic deletion of BEEs
+	AutoDelete                 *environment.AutoDelete `json:"autoDelete" form:"autoDelete"`
```
choover-broad, line 46: the key was deleting the comment

jack-r-warren (approved):

PR #117, opened 2023-01-18
[DDO-2592] Break superset relationship between readable and settable fields

jack-r-warren:
Unfortunately a huge diff, but the actual changes to the API are tiny:

- Removing Changeset POST as discussed [here](https://broadinstitute.slack.com/archives/C03TNMDJ0F8/p1674067663444429)
- ValuesName is now configurable when creating an Environment (it used to _always_ be set automatically, and it still will if left blank, but it can be configured now, because that's actually easier to implement with how defaults are handled by the controller now. Not sure what the use case for manually setting it is but it saves me a lot of code to do it this way.)

Internally, the real meat of the change is this:

```diff
- type ModelController[M v2models.Model, R Readable, C Creatable[R], E Editable[R, C]] struct {
+ type ModelController[M v2models.Model, R Readable[M], C Creatable[M], E Editable[M]] struct {
```

Rather than the controller types caring about each other, they just care about resolving down to the model. **Under the hood that doesn't really change much,** because it is basically this:

```diff
- func (c CreatableChart) toReadable() Chart {
-	return Chart{CreatableChart: c}
- }
- func (c EditableChart) toCreatable() CreatableChart {
-	return CreatableChart{EditableChart: c}
- }
+ func (c CreatableChart) toModel(storeSet *v2models.StoreSet) (v2models.Chart, error) {
+	return Chart{CreatableChart: c}.toModel(storeSet)
+ }
+ func (c EditableChart) toModel(storeSet *v2models.StoreSet) (v2models.Chart, error) {
+	return CreatableChart{EditableChart: c}.toModel(storeSet)
+ }
```

In fact, the public methods on ModelController literally didn't change their type signature at all, so I didn't need to change tests. The only change to handlers was to accept the new signature of ModelController itself.

### DIFF
--- a/internal/controllers/v2controllers/app_version.go
+++ b/internal/controllers/v2controllers/app_version.go
@@ -29,6 +29,7 @@ type EditableAppVersion struct {
 	Description string `json:"description" form:"description"` // Generally the Git commit message
 }
 
+//nolint:unused
 func (a AppVersion) toModel(storeSet *v2models.StoreSet) (v2models.AppVersion, error) {
 	var chartID uint
 	if a.Chart != "" {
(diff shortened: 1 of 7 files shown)


mflinn-broad (approved):
Makes sense to me. I like removing the self referential aspect of the generic type signatures. It's easier to understand in my opinion

PR #125, opened 2023-02-15
[DDO-2669] Add offline schedule fields

jack-r-warren:
The fields describe when the BEE should be offline. Begin is the beginning hour/minute/second that it should become offline each day and the end is the hour/minute/second that it should come back online each\* day.

\*When the EndWeekdayOnly field is enabled, the schedule won't "end" on weekends, so BEEs will sleep through the day. The "begin" side of the schedule always takes effect each day, it'll just have no effect if the BEE is already offline from the previous schedule.

I considered adding a calculated field of "should be offline right now," but that's actually harder to implement here than in Thelma where we're scanning environments, and Beehive doesn't care about it (I don't want to slow down listing all environments by also calculating a bunch of predicted offline states). I think it makes sense for Thelma to just do the work here and use Sherlock just as book-keeping.

Also removed a deprecated field that Beehive no longer uses; [`2597041`](https://github.com/broadinstitute/sherlock/pull/125/commits/259704119a509d7303e685c8a687840d1006c866)


mflinn-broad:
>  I think it makes sense for Thelma to just do the work here and use Sherlock just as book-keeping.
Amen

mflinn-broad (approved):
lgtm

PR #126, opened 2023-02-17
[DDO-2669] Rework offline schedule fields

jack-r-warren:
I don't like the design I came up with before, I like this one better. This one is a little bit more flexible and makes the Beehive logic a little bit easier. These fields haven't been in a deployed version of Sherlock yet so I'm going to do a bad thing and just kinda ignore semver or any notion of migration...


jyang-broad (approved):
:stamp:

i like the new column names

PR #131, opened 2023-03-03
[DDO-2704] DatabaseInstance as Sherlock data type

jack-r-warren:
~I skipped tests for now, wanted to get this posted so folks could check the schema and see if it makes sense before I go down that road.~ **Tests added now**

When you create an Environment from a template with autoPopulateChartReleases enabled, after creating each new ChartRelease it'll check if there was a DatabaseInstance associated to the template's ChartRelease. If so, it'll add a new copy except associated to the new ChartRelease in the BEE.

When you delete a ChartRelease, it'll delete any associated DatabaseInstance first. Deleting an Environment also deletes each ChartRelease it contains first, so the whole thing should cascade well enough, even if this'll double the number of (tiny) database queries made during the delete transaction. Not sure I see another way to do it. Plus, Thelma is what actually deletes; the user's browser is never waiting on that operation, so whatever.

To make those nested operations a bit more efficient, I slightly refactored the internalModelStore class to expose not just .get and .delete but also .getIfExists and .deleteIfExists. That cuts out a query from each iteration (or a swallowed error, depending on how you look at it).


internal/controllers/v2controllers/chart_release_test.go
```diff
@@ -537,69 +537,55 @@ func (suite *chartReleaseControllerSuite) TestChartReleaseGetOtherValidSelectors
...
-	//	suite.seedEnvironments(suite.T())
-	//	suite.seedCharts(suite.T())
-	//	suite.seedChartReleases(suite.T())
-	//
-	//	_, err := suite.ChartReleaseController.Edit(datarepoProdChartRelease.Name, EditableChartRelease{
-	//		TargetAppVersionUse: testutils.PointerTo("something obviously incorrect"),
-	//	}, auth.GenerateUser(suite.T(), true))
-	//	assert.ErrorContains(suite.T(), err, errors.BadRequest)
-	//})
+	suite.Run("successfully", func() {
```
jack-r-warren, line 540: This test had been commented from when ChartRelease didn't actually have any editable fields. It does not, so I re-enabled it and tweaked it to work with the current fields. It doesn't impact or test the new functionality here at all.

katiewelch (approved):

PR #133, opened 2023-03-07
[DDO-2709] Users in Sherlock

jack-r-warren:
This PR adds a super simple user table to Sherlock -- email + google id -- and just uses it in a passive way to basically validate that users google id doesn't change between invocations.

No, that's not very useful. The user table isn't actually exposed in the API yet and it isn't even really tested, but all the _other_ tests pass and that's what's important. There's no current need for an API here, and one will only become necessary once Sherlock does something else with users -- which will involve some model changes anyway, so the tests and API would need to be reworked.

### DIFF
--- a/db/migrations/000035_add_user_table.down.sql
+++ b/db/migrations/000035_add_user_table.down.sql
@@ -1,0 +1,1 @@
+drop table if exists v2_users;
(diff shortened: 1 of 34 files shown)


jyang-broad (approved):
looks sane. It looks like there's not currently a users table and this creates it.

PR #135, opened 2023-03-10
[DDO-2709] Sherlock GHA Auth Transparency

jack-r-warren:
This is a summary PR because I had to work from Sherlock's main branch and dev environment to test some of what's going on here.

## What's the goal?

**As an example user story:**

As a DSP or partnership developer, when I merge a PR, I'd like any triggered GitHub Action to run as "me" as much as possible, with my permissions.

**More directly:**

Sherlock should be able to assess suitability or other permissions based on the caller of the GitHub Action, not just the GCP service account used for Identity Aware Proxy authorization.

> **Wait, what?**
>
> Sherlock is behind Google's Identity Aware Proxy -- it's impossible to get a request through to Sherlock without authenticating to Google Cloud.
>
> Sherlock pays attention to the email address of the Google account used to get through IAP. Sherlock assigns permissions internally based on that email address.
>
> To contact Sherlock from GitHub Actions, we're forced to use a GCP service account to get through IAP. Up until now, this makes Sherlock unaware of who is actually making the request. This causes issues from both a usability and a security standpoint.

## Step 1: Teach Sherlock to Remember Users

When every incoming request had an email that could be mapped directly to any elevated permissions, Sherlock didn't actually need to store anything at all about users. To prepare Sherlock to remember GitHub account information for each user, we'll need to have it remember users first.

https://github.com/broadinstitute/sherlock/pull/133 implements this by adding a new table to Sherlock's database and having Sherlock's IAP-parsing code passively note users as they make requests. Since that IAP-parsing code runs in middleware before Sherlock's actual handlers receive the request, this change guarantees that any request processed by Sherlock will correlate to an exact user entry already in the database.

## Step 2: Teach Sherlock What GitHub Accounts Belong to Whom

Our users already sign in to Beehive with both IAP and GitHub authentication, but we need to move knowledge of that association from the stateless Beehive to the stateful Sherlock.

https://github.com/broadinstitute/sherlock/pull/134 implements Sherlock support for this. It exposes one API endpoint that can receive a GitHub access token. Sherlock will then call the GitHub API using that token to determine the GitHub username/ID of the requester.

In this way, Sherlock knows the association is valid: the user has proven their IAP identity by being able to make the request in the first place, and they've proven their GitHub identity by being able to pass a GitHub access token. Sherlock makes an edit to otherwise-immutable fields on the user's entry in the database to record the association.

https://github.com/broadinstitute/beehive/pull/136 has Beehive asynchronously call Sherlock's new endpoint with the user's GitHub access token as they load Beehive pages. 

> **Is this secure?**
>
> While it's theoretically possible to make a "incorrect" association, like associating your personal GitHub to a GCP service account or vice versa, such associations don't grant the user any more access than they already had (because you would've already needed access to both accounts to build the association). Sherlock prevents one GitHub account from being associated to multiple users, and users from affecting anyone else's entry in the database.

## Step 3: Teach Sherlock How To Parse GitHub Actions OIDC Tokens

GitHub Actions workflows can generate OIDC tokens [proving the identity of the workflow](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token) -- not only what workflow, but _who ran it_, according to their GitHub user ID.

Typically, those tokens get passed to cloud providers, to let workflows in a specified repository use a specified service account, for example. However, nothing is stopping us from identifying the caller of the workflow.

https://github.com/broadinstitute/sherlock/commit/b84654cf3480aea007bcb6c142aa17df7dc96c17 implements this behavior. GitHub Actions OIDC tokens are parsed from a `X-GHA-OIDC-JWT` header once IAP authentication has been processed. If the token correlates to an already-associated GitHub account (in other words, if Sherlock knows of an IAP user associated to that GitHub account), Sherlock will mark that internally-stored user as the one making the request. This makes the rest of Sherlock behave as if the request had come from the user directly. If there's no match or no token passed at all, Sherlock behaves as it did before--with the baseline-access service account being the user.

> Note that the true nature of the request isn't hidden from the rest of Sherlock -- it is readily apparent at Sherlock's permissions layer exactly how the user was authenticated for each particular request. That we currently give the same permissions regardless is an intentional choice and one we can very easily change in the future. [Here](https://github.com/broadinstitute/sherlock/actions/runs/4388657359) is an example of this feature working -- `rawInfo` is a dump of the information available to Sherlock's permissions system.
>
> Note also that any usage of this behavior is thoroughly logged as soon as the OIDC token is parsed, improving the audit trail and helping us evaluate adoption.

## Step 4 (This PR): Make All GitHub Actions Calls to Sherlock Pass OIDC Tokens

This PR implements the last step: passing tokens in the special header with all Sherlock requests. GitHub's `actions/github-script` step exposes easy access to GitHub's `@actions/core` library that in turn has [a handy `getIDToken()` function](https://github.com/actions/toolkit/tree/main/packages/core#oidc-token). It handles obtaining the token and automatically masks it from the logs. 

This PRs runs that step in all the client actions and passed the result in the header. All callers of these reusable workflows already allow OIDC tokens to be generated, because they all use OIDC-reliant Workload Identity Federation to connect to Sherlock through IAP.

That's a lot of words, but the bottom line is that no one will notice anything different except it working better. We can safely restrict more permissions (or build features that require stricter permissions) because Sherlock will see humans as running most requests, instead of Google service accounts calling from GitHub Actions.


mflinn-broad (approved):
I haven't looked at the individual PRs / commits but the summary makes sense to me. This is very impressive / cool

PR #138, opened 2023-03-17
[DDO-2741] Allow exclusion from monolith

jack-r-warren:
Chart releases now have a new editable field, `IncludeInBulkChangesets`. If explicitly set to false, it won't normally be selected in bulk updates. It can be directly included by explicitly specifying the `IncludeCharts` field on a bulk update--when you specify that field, the charts you select and only the charts you select will be bulk updated.

Beehive uses the IncludeCharts field in its UI by default, actually, so we'll need a UI pass for this to actually work there too.


internal/controllers/v2controllers/changeset_procedures.go
```diff
@@ -75,35 +75,38 @@ func (c ChangesetController) changesetPlanRequestToModelChangesets(request Chang
...
+			_, explicitlyIncluded := chartsToExplicitlyInclude[potentialTargetChartRelease.ChartID]
+
+			_, explicitlyExcluded := chartsToExplicitlyExclude[potentialTargetChartRelease.ChartID]
+
+			defaultIncluded := potentialTargetChartRelease.IncludeInBulkChangesets == nil || *potentialTargetChartRelease.IncludeInBulkChangesets
+
+			if explicitlyIncluded || (!explicitlyExcluded && defaultIncluded) {
+				targetChartReleases = append(targetChartReleases, potentialTargetChartRelease)
 			}
-			targetChartReleases = append(targetChartReleases, potentialTargetChartRelease)
```
jack-r-warren, line 106: This is the workhorse that does everything. I rewrote the logic here to make it a bit clearer.

internal/controllers/v2controllers/changeset_procedures.go
```diff
@@ -75,35 +75,38 @@ func (c ChangesetController) changesetPlanRequestToModelChangesets(request Chang
...
-			if _, excluded := chartsToExclude[potentialTargetChartRelease.ChartID]; excluded {
-				continue
+
+			_, explicitlyIncluded := chartsToExplicitlyInclude[potentialTargetChartRelease.ChartID]
+
+			_, explicitlyExcluded := chartsToExplicitlyExclude[potentialTargetChartRelease.ChartID]
+
+			defaultIncluded := potentialTargetChartRelease.IncludeInBulkChangesets == nil || *potentialTargetChartRelease.IncludeInBulkChangesets
+
+			if explicitlyIncluded || (!explicitlyExcluded && defaultIncluded) {
```
jyang-broad, line 107: Can you add a comment for how this is expected to resolve? I think this has priorities:
1) explicityIncluded 
2) explicitlyExcluded 
3) DefaultIncluded
Mostly just want to know how the explicit/ex/includes interact.

jyang-broad (approved):
looks sane!

PR #145, opened 2023-03-31
[DDO-2759] Make Environment owner field reference a User

jack-r-warren:
The older `owner` field is still around behind the scenes, now called `legacy_owner` in the database. There's a new `owner_id` field that references the user table. The controller still just exposes an `owner` field through the API (it includes `ownerInfo` if there is any, too).

Here's how it works:
- When the controller parses a controller type to a model type, it inserts goes from controller `owner` directly to model `owner_id`. There's no usage of `legacy_owner`. This does mean that `legacy_owner` can't be queried through the API now, unfortunately -- I think that's okay here because neither Beehive nor Thelma currently do that, much less rely on the stability of its output.
- The model validates that either `owner_id` or `legacy_owner` is filled. The latter can't be set once this is merged, it's just for backwards compatibility (so existing entries in the database will all still pass validation and can be edited without issue).
- When the controller parses a model type back into a controller type, it tries to use the email from the user reference in `owner_id`, and it falls back to `legacy_owner` if it that's empty. This means that existing `owner` values will still be returned from the API until a proper user reference is in `owner_id`.

There's like 30 lines of actual code here, in the first bit of https://github.com/broadinstitute/sherlock/commit/835be1032820f61fb4b42ed8f13e63ac5f887209 and https://github.com/broadinstitute/sherlock/commit/04d736035c8f892ba1d3aa61d69d4686effc6cf5. All the rest are test implications.


em-may (approved):
Description seems sane, I defer to others for code review.

PR #148, opened 2023-04-27
add new report chart version call

jyang-broad:
creates a new workflow call action for reporting chart versions to sherlock, similar to client-report-app-version.

Differences:
- parent-chart-version is now optional, as chart publishing requires a git push, and is optional on the repo, so using the git tag log is unreliable. So instead of trying to predict and possibly getting the parent wrong, just don't set it.
 - datarepo, for instance will promote `datarepo-api` or `datarepo-ui` and then `datarepo`, right now these are seperate-serial commits, but we should not depend on this behavior for correctness.
- git PR branch removed. expect that published charts are always mainline. left override in case of some wierd edge case.
- commit unchanged
- description unchanged, might end up being a trivial comment (i.e. automated chart bump), but could be useful and won't be wrong.

### DIFF
--- a/.github/workflows/client-report-chart-version.yaml
+++ b/.github/workflows/client-report-chart-version.yaml
@@ -224,6 +224,8 @@ jobs:
 
       # Parent: 
 
+      # We cannot reliably determine chart parents via git tagging because helm chart repos often have multiple charts with
+      # multiple tags so we’ll never be able to calculate parent from it without bespoke help from the caller
       - name: "Set Parent"
         if: ${{ inputs.parent-chart-version != '' }}
         shell: bash
--- a/.github/workflows/client-report-chart-version.yaml
+++ b/.github/workflows/client-report-chart-version.yaml
@@ -1,0 +1,286 @@
+name: Report Chart Version
+
+# This workflow is meant to be called from other repositories' workflows to report a new chart version to Sherlock.
+#
+# The caller repository must have Workload Identity Federation configured to allow impersonation of the
+# "dsp-tools-iap-access@dsp-tools-k8s.iam.gserviceaccount.com" service account; steps 1 and 2 of the documentation:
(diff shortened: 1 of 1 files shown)


.github/workflows/client-report-chart-version.yaml
```diff
@@ -0,0 +1,286 @@
...
+#     permissions:
+#       contents: 'read'
+#       id-token: 'write'
+# ```
+
+on:
+  workflow_call:
+
+    secrets:
+      custom-git-token:
```
knapii-developments, line 35: @jyang-broad  Can you describe a scenario where you'd want to use a custom git token?

jack-r-warren (approved):

.github/workflows/client-report-chart-version.yaml
jyang-broad: custom-git-token would be used in conjunction w/ custom-git-repo (more below)

.github/workflows/client-report-chart-version.yaml
```diff
@@ -0,0 +1,286 @@
...
+      ##   (Note that custom-git-token is available, just defined below in the secrets block)
+      ##
+
+      use-git:
+        required: false
+        type: boolean
+        default: true
+        description: "If extra version information should be gleaned from Git"
+
+      custom-git-repo:
```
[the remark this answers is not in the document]
jyang-broad: `custom-git-repo` gets called in the `checkout` action where we git clone the target repo. Usually it uses `github.repository` (which is the repo the action is running in), but `custom-git-repo` lets you call a different repository. 

The main use case for this right now is so I can have `datarepo-helm` report chart versions to sherlock, but if say, I wanted the actions to run from a different repo like `jade-data-repo` I would be setting `custom-git-repo` as well as `custom-git-token` (different permissions).

PR #149, opened 2023-05-04
add new workflow, update other dispatches

jyang-broad:
copies client-set-environment-app-version.yaml, but for charts, also updates some deprecation from using old workflow dispatch.

testing: will do testing later when calling from datarepo-helm

### DIFF
--- a/.github/workflows/client-set-environment-chart-version.yaml
+++ b/.github/workflows/client-set-environment-chart-version.yaml
@@ -150,8 +150,8 @@ jobs:
             "chartReleases": [
               {
                 "chartRelease": "${{ inputs.environment-name }}/${{ inputs.chart-name }}",
-                "toAppVersionResolver": "exact",
-                "toAppVersionExact": "${{ inputs.new-version }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/client-set-environment-chart-version.yaml
```diff
@@ -0,0 +1,256 @@
...
+
+      - name: "Write Request"
+        shell: bash
+        run: |
+          echo '{
+            "chartReleases": [
+              {
+                "chartRelease": "${{ inputs.environment-name }}/${{ inputs.chart-name }}",
+                "toAppVersionResolver": "exact",
+                "toAppVersionExact": "${{ inputs.new-version }}"
```
jack-r-warren: You'll probably need to change this

jack-r-warren (approved):

.github/workflows/client-set-environment-chart-version.yaml
jack-r-warren: Ah you just did, nvmd

PR #150, opened 2023-05-12
[DDO-2847] Make Sherlock panic if `cloudsqladmin` role is in the test database

jack-r-warren:
🤷 

I can't think of a good reason why we would ever want to run Sherlock tests against a database that had been set up by Cloud SQL... so this seems like an easy enough thing to add. It's not perfect but it's way hardier than doing nothing.

## Testing

It doesn't panic in the normal case, and then I modified the local postgres with datagrip to add a dummy role named `cloudsqladmin`. Running tests again immediately failed with 
```
test_helpers.go:20: this database looks like it is running in Cloud SQL, refusing to proceed with test harness
```
and then a giant stacktrace with a bunch of nil pointer dereferences and various panics. Close enough 🤷

### DIFF
--- a/internal/db/test_helpers.go
+++ b/internal/db/test_helpers.go
@@ -1,12 +1,28 @@
 package db
 
 import (
+	"database/sql"
+	"fmt"
 	"github.com/broadinstitute/sherlock/internal/config"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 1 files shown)


mflinn-broad (approved):
good call

PR #152, opened 2023-05-24
[DDO-2873] CiIdentifier and CiRun

jack-r-warren:
Adds CiIdentifier and CiRun types to Sherlock. Here's the database modifications.

Screenshot 2023-05-25 at 12 48 48 PM

The trick here is we're relying on Gorm to grab the appropriate CiIdentifier when one exists for another type. When Gorm reads a v2models.Chart, for example, it'll also query the v2_ci_identifiers table for resource_type = "chart" and resource_id equal to the chart's ID. If there's a match, it'll include it in the returned v2models.Chart. I made a composite index that should help speed that up.

It _is_ possible to create a CI identifier directly, there's API methods for it. But you probably won't ever need to do that.

The CiRun data type is really what does the magic. When you create or edit a CiRun, you can include selectors of other types to associate that run to. If any of those types don't already have CiIdentifiers, _Sherlock will automatically create them just-in-time for your request to go through_.

So the write flow here is "GitHub Action runs and reports info to Sherlock. Any data types it affects will have CiIdentifiers silently created for them." Meanwhile the read flow is "Beehive gets a data type, and if it contains a CiIdentifier, Beehive knows it can query that to list out any CiRuns for that data type." (I'm using data type to mean singular instances here, like a single Chart--you get the gist).

Sherlock will get super smart with this, too. If a GitHub Action says it affects a Changeset, Sherlock is going to record it as affecting:
- that Changeset
- any new App Versions the Changeset is deploying (it's that changelog gimmick again, so it'll pick up intermediate versions)
- ditto for Chart Versions
- the Chart Release being affected
- the Environment the Chart Release is in, if there is one
- the Cluster the Chart Release is in, if there is one

I've got a test written for precisely this behavior, because that's like 90% of how we're probably gonna populate this CiIdentifier table and get stuff showing up in Beehive.

## Testing

Speaking of testing, here's [the big test](https://github.com/broadinstitute/sherlock/blob/DDO-2873-ci/internal/controllers/v2controllers/ci_run_test.go). If there's part of this that anyone's gonna look at, I think it should be this part. I think the test file is the best documentation of how everything works together.

There's also a ton of unit tests for all the validation code etc

## Risk

So the API design here was carefully done so that existing API responses don't balloon in size. There's no pagination if you load a CiIdentifier or CiRun directly, [so we may need to address that down the line](https://broadinstitute.slack.com/archives/CQ6SL4N5T/p1685025006984109), but it won't cause a regression in existing behavior if that becomes an issue. Otherwise I think the risk is pretty low--will deploy to dev before prod to make sure.

### DIFF
--- a/db/migrations/000042_create_ci.down.sql
+++ b/db/migrations/000042_create_ci.down.sql
@@ -1,0 +1,5 @@
+drop table if exists v2_ci_runs_for_identifiers cascade;
+
+drop table if exists v2_ci_identifiers cascade;
+
+drop table if exists v2_ci_runs cascade;
--- a/db/migrations/000042_create_ci.up.sql
+++ b/db/migrations/000042_create_ci.up.sql
@@ -1,0 +1,52 @@
+create table v2_ci_identifiers
+(
+    id            bigserial
+        primary key,
+    created_at    timestamp with time zone,
+    updated_at    timestamp with time zone,
+    deleted_at    timestamp with time zone,
+    resource_type text,
+    resource_id   bigint
+);
+
+create index idx_v2_ci_identifiers_deleted_at
+    on v2_ci_identifiers (deleted_at);
+
+create index idx_v2_ci_identifiers_polymorphic_index
+    on v2_ci_identifiers (resource_type, resource_id);
+
+create table v2_ci_runs
+(
+    id                            bigserial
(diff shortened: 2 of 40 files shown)


choover-broad (approved):
💪

PR #167, opened 2023-06-06
[DDO-2896] Fix Sherlock metrics

jack-r-warren:
I totally forgot that these need to be exposed by Sherlock's actual router. They were being exposed by the v1 api part so I forgot that they were still necessary here.

## Testing

404'ed before locally, now it returns normally

## Risk

Super low, this is a bug fix just exposing existing functionality


mflinn-broad (approved):

jyang-broad (approved):
:stamp:

PR #176, opened 2023-06-09
[DDO-2900] Sherlock Webhook Proxy

jack-r-warren:
It's a bird, 
It's a plane, 
It's a Go cloud function that sits publicly at https://sherlock-webhook-proxy-6f6srdhl6q-uc.a.run.app and turns GitHub webhook payloads delivered to /webhook into authenticated Sherlock API calls.

A bit more specifically, it turns GitHub [workflow_run payloads](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run) into Sherlock [CiRun PUT requests](https://sherlock.dsp-devops.broadinstitute.org/swagger/index.html#/CiRuns/put_api_v2_ci_runs__selector_).

How is that safe? It verifies the signature sent along with the payload, making sure that it was sent with [our secret](https://console.cloud.google.com/security/secret-manager/secret/sherlock-prod-webhook-secret/versions?project=dsp-tools-k8s) that we [created with Terraform](https://github.com/broadinstitute/terraform-ap-deployments/blob/master/sherlock/webook_proxy.tf#L7) (the cloud function can [read it from Google Secret Manager natively](https://github.com/broadinstitute/sherlock/pull/176/files#diff-beda42571c095172ab63437d050612a571d0d9ddd3ad4f2aecbce907a9b7e3d0R53), while we use Terraform again to [add it to the webhook configuration](https://github.com/broadinstitute/terraform-ap-deployments/blob/master/github/webhook.tf#L20)). The cloud function's own service account has basic IAP access, just enough to [authenticate to Sherlock](https://beehive.dsp-devops.broadinstitute.org/users/sherlock-webhook-proxy@dsp-tools-k8s.iam.gserviceaccount.com).

> **https://sherlock-webhook-proxy-6f6srdhl6q-uc.a.run.app? That's a weird URL...**
>
> Yeah, Gen 2 cloud functions have stable URLs (better than v1, actually) but they look a bit weird. No matter, we [read it back out of GCP directly to add to our webhook config](https://github.com/broadinstitute/terraform-ap-deployments/blob/master/github/webhook.tf#L18).

> **How sure are we that the request comes from GitHub?**
>
> GitHub sends signatures in two different headers, a worse one (HMAC + SHA-1) and a better one (HMAC + SHA-256). We're using a library that [automatically checks the worse one](https://github.com/go-playground/webhooks/blob/master/github/github.go#L167), but I didn't take a crypto class for nothing so we [play a little trick with `io.TeeReader`](https://github.com/broadinstitute/sherlock/pull/176/files#diff-e60f4d1e3988e6313a5435f8095b552e922ca1507de23fbfb1a3c4dc20fe1defR100) to let us also [check the better one](https://github.com/broadinstitute/sherlock/pull/176/files#diff-e60f4d1e3988e6313a5435f8095b552e922ca1507de23fbfb1a3c4dc20fe1defR138).

> **What about the deployment process? Are we not Terraforming this cloud function?**
>
> No, I didn't want to. We Terraform the secret and IAM but Terraforming the actual cloud function itself means we'd need to delve into Cloud Build / Cloud Source Repositories / etc to get automatic deploy upon merge working. Frankly, I think a [simple GHA](https://github.com/broadinstitute/sherlock/pull/176/files#diff-1c89de5fdc3d507eff77f346d192e6add9dbb99ebb0018ea615ad381d9147b3c) that just calls [the one `gcloud` command to do it](https://github.com/broadinstitute/sherlock/pull/176/files#diff-beda42571c095172ab63437d050612a571d0d9ddd3ad4f2aecbce907a9b7e3d0R43) is easier to maintain, because it's right here.

> **So the function is just one big file?**
>
> Yeah... it's like 200 lines, and it just runs linearly from start to finish, might as well be a bash script. But we absolutely can make it better, I just wanted to get a MVP out the door. I think when we eventually want to collect some other webhook payloads we'll have a natural point to improve the code here.

## Testing

It's deployed and working -- you can see the webhook config on repos [like this one](https://github.com/broadinstitute/sherlock/settings/hooks/418574304), you can see the cloud function [invocations](https://console.cloud.google.com/functions/details/us-central1/sherlock-webhook-proxy?env=gen2&project=dsp-tools-k8s) and [logs](https://console.cloud.google.com/functions/details/us-central1/sherlock-webhook-proxy?env=gen2&project=dsp-tools-k8s&tab=logs), and if you connect to Sherlock's database you can check the v2_ci_runs table (there is an API but it doesn't paginate well yet so I don't suggest using it to get a sense of all the data being collected, it's a lot).

[I tested the GitHub Action on the first commit on this branch](https://github.com/broadinstitute/sherlock/actions/runs/5216861634/jobs/9416035005), in the future it'll run just upon merges.

## Risk

This PR is already deployed, it's new capability nothing is/was relying on, so this is super low risk IMO.


sherlock-webhook-proxy/pkg/handler.go
```diff
@@ -0,0 +1,248 @@
...
+		// Require that the signature header is present
+		signature := strings.TrimPrefix(r.Header.Get("X-Hub-Signature-256"), "sha256=")
+		if signature == "" {
+			w.WriteHeader(http.StatusUnauthorized)
+			log.Printf("request was missing X-Hub-Signature-256 header\n")
+			return
+		}
+
+		// As r.Body is read, additionally synchronously write it into _mac
+		_mac.Reset()
```
choover-broad, line 101: does cloud functions guarantee that only one request is handled at a time? otherwise keeping this in a package-level variable doesn't seem concurrency safe 🤔

sherlock-webhook-proxy/pkg/handler.go
```diff
@@ -0,0 +1,248 @@
...
+		r.Body = struct {
+			io.Reader
```
choover-broad, line 103: this is a very neat way to wrap a ReadCloser!

sherlock-webhook-proxy/pkg/handler.go
```diff
@@ -0,0 +1,248 @@
...
+			var status, terminalAt string
+			if payload.WorkflowRun.Conclusion != "" {
+				status = payload.WorkflowRun.Conclusion
+				terminalAt = payload.WorkflowRun.UpdatedAt.Format(time.RFC3339)
+			} else {
+				status = payload.WorkflowRun.Status
+			}
+
+			// PUT to Sherlock; will create if the selector isn't found and edit otherwise
+			edited, created, err := sherlockClient.CiRuns.PutAPIV2CiRunsSelector(&ci_runs.PutAPIV2CiRunsSelectorParams{
```
choover-broad, line 200: if there's ultimately still a 1:1 ratio between GitHub requests and requests to sherlock, I guess the advantage of having this in a cloud function is that:

(1) we reject invalid/unauthed requests before they even reach Sherlock, which protects from abuse
(2) we don't have to poke a hole in IAP, which protects us from accidentally making that hole bigger than it should be

is that correct?

choover-broad (approved):

sherlock-webhook-proxy/pkg/handler.go
jack-r-warren: It does! With cloud functions gen2 it would be possible for us to override that, to make it handle requests concurrently, but by default they're one at a time. It does keep instances alive, which is why I'm using `init` and package-level stuff to make that run faster after cold-start

sherlock-webhook-proxy/pkg/handler.go
jack-r-warren: Yeah, my whole idea for serverless here is that it makes this a lower risk change, because it doesn't actually change Sherlock's security model. Sherlock's IAP isn't any less effective at protecting it after this change. My other idea is that it's super easy to modify this cloud function to handle other payloads, that potentially don't have that 1:1 relationship (for example, could subscribe to `issue_comment` payloads and increment a per-repo counter in sherlock whenever `jenkins retest` is commented)

PR #178, opened 2023-06-09
[DDO-2903] Add a startedAt field to the CiRun schema

jack-r-warren:
There's a createdAt that I was intending to use, but that actually gets set when the workflow is queued, as opposed to when it starts running (based on how GitHub sends up webhooks). Figure I'll make space to record the actual value and make a subsequent PR once this is deployed to update the cloud function.

## Testing

Migration worked against my copy of the database, and I added tests

## Risk

Minimal IMO


jyang-broad (approved):
:stamp:
adds a new started_at column.
also tests it.

PR #202, opened 2023-07-07
[DDO-2944] More refactoring

jack-r-warren:
#### The usual PR stuff

Why all of this? To add some query parameters to an endpoint. Yes, really.

Testing? This PR has super thorough testing, so much that it bumps project coverage by nearly 7%. More info below on why that's an understatement. (I'm not trying to be cocky, just trying to explain why this PR took so long ☹️)

Risk? Pretty low actually, I think. [I've already removed Beehive's dependency on the one removed endpoint](https://github.com/broadinstitute/beehive/pull/251), and you can check the endpoint diff below for evidence that nothing else is being impacted on the current APIs. I've tested the additional checks being added to the DB against a copy of the prod database (the checks were already being done, just at the application layer).

## The big refactor

The old way was something like this:

1. Gin called handlers,
2. handlers exclusively called controllers,
3. controllers exclusively called models,
4. models exclusively called Gorm

The new way is like this:

1. Gin calls handlers
2. handlers call Gorm _with data types from the models package_

The models package defines the database types and any Gorm hooks (BeforeCreate, AfterFind, etc) but it doesn't wrap Gorm.

This means we don't have to write nearly as much code, while simultaneously being able to bolt on extras where we need.

For example, if we want to add pagination parameters to an endpoint... [_we can just do that_](https://github.com/broadinstitute/sherlock/pull/202/commits/c0dcb715c43fc7ad328f0f008aad9a605e7ba2eb#diff-2828e228d03c0a95c1d18ece5cbe10773444071bae9f2661b57d8c15262f722e). Thank goodness.

> **Note**
>
> **Wait, the new API code / API handlers are in a package called... `sherlock`?**
>
> Yeah, I'm not terribly happy about it either, but here's the deal. The Swagger spec picks up the package name that your API stuff lives in and uses it _everywhere_. This is why there's about a billion references to "v2controllers" inside Beehive's codebase: everything from the client library is prefixed like "v2controllersUser".
>
> Just biting the bullet and calling the package `sherlock` (full path is sherlock/internal/apis/sherlock, next to sherlock/internal/apis/misc for the status and version endpoints) means that we'll get satisfying types like "sherlockUser" and "sherlockEnvironment" inside the client libraries.

### Fixing my own tech debt with `auth_models.User`

Most of Sherlock's understanding of users predates it actually storing users as a CRUD data type -- it always needed a magic type to hold the caller's identity, and that being persisted in the database was bolted on later. Rather than coupling the big refactor to my past mistakes, I moved the magic type to the big refactor (`auth_models.User` -> `models.User`).

In other words, the big refactor has CiRun/CiIdentifier at the API level, but CiRun/CiIdentifier/**_User_** at the model level. The old code still gets to treat like it an opaque non-CRUD type, but it's fully native to the new code. We'll need to add the refactored User to the refactored API but that should be a small lift since it is API-read-only.

Why does this matter at all? This means new code at the API layer can actually call old controllers/models without needing to bend over backwards or getting icky circular dependencies.

> **Note**
>
> How do package dependencies work between the old code and the new code? Every old code package now depends on the new models, but the new API code can freely depend on every old code package in addition to the new models.
>
> This isn't circular because the old code doesn't need to call the new API code.
>
> This does mean that as we add more new API code, we'll find it helpful to go in database dependency order (i.e. charts before things that reference charts). I don't see that as a bad thing.

#### Validation now done by SQL

Validation and uniqueness checks are being added to the database for refactored data types. We still unit test this behavior from Go, but applying it from SQL helps us avoid wrapping Gorm.

> **Warning**
>
> Yes, we do now have to worry about uniqueness not tripping on soft-deletes, which has been an issue in the past (#92).
>
> Postgres to the rescue: it has wonderful support for "partial unique indexes" which does exactly what we want.

#### `auth` -> `authentication` and `authorization`

The poorly-named `auth` package has been split up. Both `authentication` and `authorization` now formally support local and test users, meaning that offline Sherlock now runs much more similarly to when it's deployed.

#### Test coverage

The codecov report understates the jump in test coverage; things like application start/stop, readiness/liveness endpoints, embedded config/migration files, and more are evaluated by the test suite, even if they're the responsibility of a library.

Combined with how we're testing the new API surface, this is huge -- we're basically going from "testing the code we wrote" to "testing that the app works." This is especially helpful for feeling confident in merging Dependabot PRs.

#### Performance

I know we don't care so I didn't collect numbers... but in some places this new design is cutting down on database operations by an order of magnitude (using Gorm directly lets us farm more things out to SQL). This does matter for bulk updates from Beehive, which can take several seconds currently.

The test harnesses are faster, too: the entire new API+model surface runs its full suite in a second and a half.


mflinn-broad (approved):

choover-broad (approved):
sounds like there are a lot of wins with this refactor :tada: -- looking forward to the demo!

PR #212, opened 2023-07-25
[DDO-2981] Improve app/chart version CI reporting based on changesets

jack-r-warren:
The old CiRun PUT endpoint had this gimmick where when you report that a GHA relates to a Changeset, it also records that the GHA relates to any new app or chart versions that Changeset deployed.

That works fine, but it would get spammy if we use it more, because BEE deployments would then show up when you look at an app or chart version's CiRuns.

So the v3 one I added has that behavior behind a flag in the request. It's disabled by default. No more spam.

Except... what we really want is "enable this for deployments to static environments, disable it otherwise." We _want_ deployments to dev/alpha/staging/prod to show up if you look at the CiRuns for an app version.

Rather than requiring clients to figure that out and control the flag, this PR turns the boolean flag into an enum that accepts `always`, `when-static`, and `never`. `when-static` is the default and Sherlock figures it out for you.

## Testing

Added tests for validation of the enum and for its behavior across static and non-static environments. Because I split the code paths for changeset and chart release handling by this endpoint, I also added another test for the chart release part so we aren't losing coverage we used to have.

## Risk

None, this endpoint field isn't currently being used by anyone/anything.

### DIFF
--- a/sherlock/go.mod
+++ b/sherlock/go.mod
@@ -53,7 +53,7 @@ require (
 	github.com/go-openapi/swag v0.22.3 // indirect
 	github.com/go-playground/locales v0.14.1 // indirect
 	github.com/go-playground/universal-translator v0.18.1 // indirect
-	github.com/go-playground/validator/v10 v10.14.0 // indirect
+	github.com/go-playground/validator/v10 v10.14.1 // indirect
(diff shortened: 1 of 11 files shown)


mflinn-broad (approved):
Good call, this makes more sense to me as a default behavior

PR #215, opened 2023-07-25
DDO-2849 Record deployments to pact broker

knapii-developments:
<h4> Summary </h4>

* Implementing [step 7 ](https://docs.pact.io/pact_nirvana/step_7)to pact nirvana which allows us to record deployments to pact. 
* Adds 2 data fields. PactParticipant in charts and ParticipatesInPact in environments.


jack-r-warren:
One other comment is just a heads up that this PR doesn't expose the field through Sherlock's API at all. I'm assuming that's intentional and we'll do another change to expose it (but if not, happy to pair on that).

sherlock/internal/deprecated_models/v2models/changeset.go
```diff
@@ -254,6 +256,14 @@ func (s *internalChangesetStore) apply(db *gorm.DB, changesets []Changeset, user
 					fmt.Sprintf(config.Config.MustString("beehive.chartReleaseUrlFormatString"), chartRelease.Name),
 				)
 			}
+			if chartRelease.Environment.ParticipatesInPact != nil && *chartRelease.Environment.ParticipatesInPact && chartRelease.Chart.PactParticipant != nil && *chartRelease.Chart.PactParticipant {
```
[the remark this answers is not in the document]
jack-r-warren: No, not necessarily. `*Environment` will be nil if the Gorm call that loads the chart release doesn't include `.Association("Environment")`. It's sorta just something all across Sherlock -- we always have to check nil pointers, we can't have Sherlock segfaulting at runtime.

sherlock/internal/deprecated_controllers/v2controllers/environment_test.go
```diff
@@ -273,6 +273,7 @@ func (suite *environmentControllerSuite) TestEnvironmentCreate() {
 			EditableEnvironment: EditableEnvironment{
 				DefaultCluster: &terraDevCluster.Name,
 				Owner:          testutils.PointerTo("some-email-not-in-the-database@broadinstitute.ogr"),
+				PactIdentifier: nil,
```
jack-r-warren, line 276:
```suggestion
```

Setting this field to nil is a no-op, it'll be nil if unset. To test your controller/api-level changes, I'd suggest adding to the edit tests in `environment_test.go` and `chart_test.go`; I think the methods will be called `TestEnvironmentEdit()` and `TestChartEdit()`.

sherlock/internal/deprecated_models/v2models/changeset.go
```diff
@@ -254,18 +256,29 @@ func (s *internalChangesetStore) apply(db *gorm.DB, changesets []Changeset, user
...
+			for environmentID, chartReleaseNames := range environmentReleases {
+				environment, err := InternalEnvironmentStore.Get(db, Environment{Model: gorm.Model{ID: environmentID}})
+				if err == nil && environment.PagerdutyIntegration != nil && environment.PagerdutyIntegration.Key != nil {
+					go pagerduty.SendChangeSwallowErrors(
+						*environment.PagerdutyIntegration.Key,
+						fmt.Sprintf("Version changes to %s via Sherlock/Beehive", strings.Join(chartReleaseNames, ", ")),
+						fmt.Sprintf(config.Config.MustString("beehive.environmentUrlFormatString"), environment.Name),
+					)
+				}
+			}
```
jack-r-warren, line 269: It looks like your edits accidentally nested this `range environmentReleases` loop inside the `range affectedChartReleases` loop. It should go after the `range affectedChartReleases`, not inside it.

sherlock/internal/pactbroker/record_deployment.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"errors"
+	"fmt"
+	"github.com/broadinstitute/sherlock/sherlock/internal/config"
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+	"net/http"
+)
+
+const pburl = `https://pact-broker.dsp-eng-tools.broadinstitute.org`
```
jack-r-warren, line 12: I think this string should go in default_config.yaml, we shouldn't be hardcoding URLs as consts inside the codebase

sherlock/internal/pactbroker/record_deployment.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Record deployment to pact broker
+// https://docs.pact.io/pact_broker/recording_deployments_and_releases
+func RecordDeployment(chartName string, appVersion string, eID uuid.UUID) {
+	if config.Config.Bool("pactbroker.enabled") {
+
+		request, err := http.NewRequest(http.MethodPost, pburl+"/pacticipants/"+chartName+
+			"/versions/"+appVersion+"/deployed-versions/environment/"+eID.String(), nil)
+		if err != nil {
+			PactSwallowErrors(err)
```
jack-r-warren, line 22:
```suggestion
			return swallowErrors(err)
```

sherlock/internal/pactbroker/record_deployment.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		request.Header.Set("Content-Type", "application/json; charset=utf-8")
+		request.Header.Set("Accept", "application/hal+json")
+		request.SetBasicAuth(config.Config.MustString("pactbroker.auth.username"), config.Config.MustString("pactbroker.auth.password"))
+		// send the request
+		client := &http.Client{}
+		response, err := client.Do(request)
+		if err != nil {
+			PactSwallowErrors(err)
```
jack-r-warren, line 31:
```suggestion
			return swallowError(err)
```

sherlock/internal/pactbroker/record_deployment.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+		if response.StatusCode != 201 {
+			errMsg := fmt.Sprintf("Deployment for %s app version %s was not recorded to pact successfully.\n"+
+				"Return code %d ", chartName, appVersion, response.StatusCode)
+			PactSwallowErrors(errors.New(errMsg))
```
jack-r-warren, line 36:
```suggestion
			return swallowError(fmt.Errorf("deployment for %s app version %s was not recorded to pact successfully (return code %d)", 
				chartName, appVersion, response.StatusCode))
```
`fmt.Errorf` is basically `errors.New(fmt.Sprintf)`. Go also has some soft standards around errors (basically, that they start with a lower case letter, don't end with a period, and are one line)

sherlock/internal/pactbroker/record_deployment.go
```diff
@@ -0,0 +1,43 @@
...
+		}
+	}
+}
+
+func PactSwallowErrors(err error) {
+	log.Warn().Msgf("%v", err)
+}
```
jack-r-warren, line 43:
```suggestion
func swallowError(err error) {
	log.Warn().Msgf("PACT | %v", err)
}
```
No need to export this function outside this package, it can be lowercase so it's just internal. Prefixing a log line like this with "PACT | " will also make it fit with all other Sherlock log lines.

sherlock/internal/pactbroker/record_deployment.go
```diff
@@ -0,0 +1,43 @@
...
+	if config.Config.Bool("pactbroker.enabled") {
+
+		request, err := http.NewRequest(http.MethodPost, pburl+"/pacticipants/"+chartName+
+			"/versions/"+appVersion+"/deployed-versions/environment/"+eID.String(), nil)
+		if err != nil {
+			PactSwallowErrors(err)
+		}
+		request.Header.Set("Content-Type", "application/json; charset=utf-8")
+		request.Header.Set("Accept", "application/hal+json")
+		request.SetBasicAuth(config.Config.MustString("pactbroker.auth.username"), config.Config.MustString("pactbroker.auth.password"))
```
jack-r-warren, line 26:
```suggestion
		request.SetBasicAuth(config.Config.String("pactbroker.auth.username"), config.Config.String("pactbroker.auth.password"))
```
I don't think this function should use `MustString`, because that will panic and crash the whole request if the config value isn't present. `String` will just result in an empty string, so we should get an authentication error -- which will probably be more helpful than a panic anyway.

sherlock/internal/pactbroker/record_deployment.go
jack-r-warren: I think when we swallow an error we should exit, instead of keep going with the rest of the function.

sherlock/internal/pactbroker/record_deployment.go
knapii-developments: bb59eba

sherlock/internal/deprecated_models/v2models/changeset.go
knapii-developments: Can [affectedChartReleases](https://github.com/broadinstitute/sherlock/blob/27b38bac7dd272e00d90dffed52837a1cbd81ea5/sherlock/internal/deprecated_models/v2models/changeset.go#L167C2-L167C23) have different charts = different services? If so, I want to record each to pact.

sherlock/internal/deprecated_controllers/v2controllers/environment_test.go
knapii-developments: In ```testEnvironment```, was your goal to test edits in every environment and a suitable environment?

sherlock/internal/deprecated_models/v2models/changeset.go
jack-r-warren: Affected chart releases is a list of each chart release that underwent a version change. I know you want to record each of those version changes to pact, that's totally fine. 

Here, it seems like you're modifying an unrelated part, the "per each environment report to pagerduty" code. I think the way you're changing it will make it not work so I think it would be better to leave it as-is.

sherlock/internal/deprecated_controllers/v2controllers/environment_test.go
knapii-developments: 1fff591

sherlock/internal/deprecated_models/v2models/changeset.go
knapii-developments: fe8c12d

knapii-developments: fe8c12d

knapii-developments: I'm not sure I understand this. Outer loop iterates through each affected chart release. For each iteration, I report to pact is an isolated if statement outside of the ```for environmentID, chartReleaseNames := range environmentReleases`` loop.

knapii-developments: a16498a

jack-r-warren: Sorry I didn't see your comment -- your commit doesn't address the concern I'm raising. Can you see how your PR indents the `for environmentID, chartReleaseNames := range environmentReleases {` loop? It moves it inside the `for _, chartRelease := range affectedChartReleases {` loop when it wasn't there before.

Screenshot 2023-08-14 at 11 12 10 AM

knapii-developments: I got it : 774c032

jack-r-warren (approved):
LGTM! Nice work

PR #227, opened 2023-08-01
[DDO-3011] v3 Chart model

jack-r-warren:
Not exposed in the API yet, just taking baby steps.

## Testing

Tests added, including for the added SQL checks.

## Risk

None


mflinn-broad (approved):

sherlock/internal/deprecated_controllers/v2controllers/chart_test.go
```diff
@@ -290,7 +290,7 @@ func (suite *chartControllerSuite) TestChartEdit() {
 		suite.seedCharts(suite.T(), suite.db)
 
 		_, err := suite.ChartController.Edit(yaleLibChart.Name, EditableChart{ChartRepo: testutils.PointerTo("")}, generateUser(suite.T(), suite.db, false))
-		assert.ErrorContains(suite.T(), err, errors.BadRequest)
+		assert.ErrorContains(suite.T(), err, "violates check constraint")
```
jack-r-warren, line 293: this test now fails at an earlier point because of the added SQL check; the `errors.BadRequest` will still get added at a higher point by the handler

PR #245, opened 2023-08-15
[DDO-3039] Set max idle connections to a high value to make Sherlock less aggressive with the cloud sql proxy

jack-r-warren:
https://broadinstitute.slack.com/archives/CQ6SL4N5T/p1692116090451579


mflinn-broad (approved):
Stamp  but consider setting https://pkg.go.dev/database/sql#DB.SetConnMaxIdleTime

PR #248, opened 2023-08-15
[DDO-3040] Update test-auth-transparency.yaml

jack-r-warren:
I only use this as a dumb manual action to run but should make it hit the right endpoints.

## Testing

https://github.com/broadinstitute/sherlock/actions/runs/5870301762 it works

## Risk

Not actually Sherlock, no risk


jyang-broad (approved):
:Stamp:

PR #275, opened 2023-08-25
[DDO-3075] Expose Sherlock's knowledge of repos and workflow files

jack-r-warren:
This will be really helpful for Beehive.

It returns data like this:

```json
{
  "CancerDataAggregator": {
    "cda-service": [
      ".github/workflows/pr_push.yml",
      ".github/workflows/tests.yml",
      ".github/workflows/trivy.yaml",
      ".github/workflows/master_push.yml"
    ]
  },
  "DataBiosphere": {
    "biocore-data-model": [
      ".github/workflows/build.yaml"
    ],
    "calhoun": [
      ".github/workflows/unit-test.yml",
      ".github/workflows/trivy.yml",
      ".github/workflows/build.yml"
    ],
```

## Testing

Added tests

## Risk

None

### DIFF
--- a/sherlock/internal/api/sherlock/ci_runs_procedures_v3_github_info_list.go
+++ b/sherlock/internal/api/sherlock/ci_runs_procedures_v3_github_info_list.go
@@ -1,0 +1,50 @@
+package sherlock
+
+import (
+	"github.com/broadinstitute/sherlock/sherlock/internal/authentication"
+	"github.com/broadinstitute/sherlock/sherlock/internal/errors"
+	"github.com/broadinstitute/sherlock/sherlock/internal/models"
+	"github.com/gin-gonic/gin"
+	"net/http"
+)
+
+// ciRunsProceduresV3GithubInfoList godoc
+//
+//	@summary		List GitHub info gleaned from CiRuns
+//	@description	List info about GitHub repos and their workflow files as determined by CiRuns from the past 90 days.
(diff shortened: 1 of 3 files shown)


sherlock/internal/api/sherlock/ci_runs_procedures_v3_github_info_list.go
```diff
@@ -0,0 +1,50 @@
...
+		return
+	}
+	type result struct {
+		GithubActionsOwner        string
+		GithubActionsRepo         string
+		GithubActionsWorkflowPath string
+	}
+	var results []result
+	if err = db.Model(&models.CiRun{}).
+		Where("created_at >= current_timestamp - '90 days'::interval").
```
jyang-broad, line 34: What's the reasoning behind the 90 days setting? Is it linked to some technical limitation or "reasonable default" kind of number.

jyang-broad (approved):
Looks sane!
Appears to add a new API.
Test checks that json returns as expected

sherlock/internal/api/sherlock/ci_runs_procedures_v3_github_info_list.go
jack-r-warren: Arbitrary balance between "if someone deletes or renames a workflow file, it shouldn't be in this endpoint's response forever" and "workflow files should be suggested even if they haven't run recently".

Making this configurable would be a SQL injection vuln and/or require an annoying amount of logic and it just doesn't really matter so I figured it was fine to hardcode

PR #317, opened 2023-10-03
[DDO-3195] Properly handle the outcome of CiRuns for conditionally triggering deploy hooks

jack-r-warren:
It was ignoring the status when deciding when to run a deploy hook before.

## Testing

Added a big ol' test that checks it now.

## Risk

Very low


choover-broad (approved):
Thanks!

PR #320, opened 2023-10-06
[DDO-3204] Changelog capability for v3 app versions

jack-r-warren:
Adds `/api/app-versions/procedures/v3/changelog`. It runs with hundreds fewer database roundtrips than `/api/v2/procedures/app-versions/children-path-to-parent` but runs the same algorithm, it just does so within the database using `WITH RECURSIVE`. We discussed this and its tradeoffs [here](https://broadinstitute.slack.com/archives/CQ6SL4N5T/p1696536340273929).

## Testing

Added tests, enough that I actually found some edge cases in the old endpoint's behavior. Rather than making breaking changes there I'll just have this endpoint handle them.

## Risk

Read-only endpoint, very low.

### DIFF
--- a/sherlock/internal/models/app_version.go
+++ b/sherlock/internal/models/app_version.go
@@ -1,6 +1,9 @@
 package models
 
-import "gorm.io/gorm"
+import (
+	"github.com/broadinstitute/sherlock/go-shared/pkg/utils"
+	"gorm.io/gorm"
+)
 
 type AppVersion struct {
 	gorm.Model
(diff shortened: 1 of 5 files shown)


choover-broad (approved):

sherlock/internal/models/app_version.go
```diff
@@ -26,3 +29,60 @@ func (a AppVersion) GetCiIdentifier() CiIdentifier {
...
+		return []uint{}, true, nil
+	}
+
+	type AppVersionPathEntry struct {
+		ID                 uint
+		ParentAppVersionID uint
+	}
+	var appVersionPath []AppVersionPathEntry
+	if err = db.Raw(
+		//language=SQL
```
jack-r-warren, line 50: Explicitly tell GoLand or other JetBrains IDEs that this string is SQL (you'll get validation directly against Sherlock's database if you've connected to it from GoLand)

PR #321, opened 2023-10-06
DDO-3201: chart version v3 API

katiewelch:


sherlock/internal/api/sherlock/chart_version_v3.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	var chart *ChartV3
+	var chartName string
+	if model.Chart != nil {
+		chart = utils.PointerTo(chartFromModel(*model.Chart))
+		chartName = chart.Name
+	}
+	var parentChartVersion *ChartVersionV3
+	var parentChartVersionString string
+	println("OUTSIDE")
```
jack-r-warren, line 84:
```suggestion
```

sherlock/internal/api/sherlock/chart_version_v3.go
```diff
@@ -0,0 +1,110 @@
...
+	if model.ParentChartVersion != nil {
+		println("INSIDE")
```
jack-r-warren, line 86:
```suggestion
```

sherlock/internal/api/sherlock/chart_version_v3_list.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		errors.AbortRequest(ctx, fmt.Errorf("(%s) %v", errors.BadRequest, err))
+		return
+	}
+	var results []models.ChartVersion
+	if err = db.
+		Where(&modelFilter).
+		Limit(limit).
+		Offset(offset).
+		Order("id asc").
```
jack-r-warren, line 56:
```suggestion
		Order("created_at desc").
```

I think we actually probably want to change this for https://github.com/broadinstitute/sherlock/blob/main/sherlock/internal/api/sherlock/app_version_v3_list.go#L56 too. The IDs are a Sherlock-internal thing so they're pretty meaningless to sort by, and I think Beehive does stuff like `/api/v2/chart-versions?limit=10` to get the ten most recent app versions, that kind of thing.

sherlock/db/migrations/000058_chart_version_v3.down.sql
```diff
@@ -0,0 +1,4 @@
+DROP INDEX IF EXISTS chart_versions_selector_unique_constraint;
+
+ALTER TABLE v2_chart_versions
+    DROP CONSTRAINT IF EXISTS chart_version_present;
```
jack-r-warren, line 4:
```suggestion
    DROP CONSTRAINT IF EXISTS chart_version_present;

```
This is what's actually causing the test failure, there's a unit test that validates that these files have newlines at the end

sherlock/db/migrations/000058_chart_version_v3.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE UNIQUE INDEX chart_versions_selector_unique_constraint
+ON v2_chart_versions (chart_id, chart_version)
+WHERE deleted_at IS NULL;
+
+ALTER TABLE v2_chart_versions
+ADD CONSTRAINT chart_version_present
+CHECK (chart_version IS NOT NULL AND chart_version != '');
```
jack-r-warren, line 7:
```suggestion
CHECK (chart_version IS NOT NULL AND chart_version != '');

```

jack-r-warren (approved):
Great stuff!

PR #327, opened 2023-10-12
[DDO-3215] Fix v3 version upsert paths

jack-r-warren:
The v3 upsert endpoints work properly but the comments were indicating the endpoints were at `/api/appVersions/v3` and `/api/chartVersions/v3` instead of `/api/app-versions/v3` and `/api/chart-versions/v3`.

This means that the client libraries were getting a gin fallback 404, instead of an API-level 404 like the client libraries are built to handle. [This is why Thelma was complaining about not being able to parse the error response](https://broadinstitute.slack.com/archives/CQ6SL4N5T/p1697048156679049?thread_ts=1697047954.711599&cid=CQ6SL4N5T) (it wasn't json that was coming back from gin).

## Testing 

I added a unit test to scan Sherlock's own source code files for declarations for API paths with capitalized letters in them. We can improve the regex over time if we need but it should help avoid this sort of thing in the future (it's super easy to miss, since it's just a comment).

## Risk

No one could be successfully using the existing endpoint because it's at an unexpected URL, so I think this is safe

### DIFF
--- a/sherlock/internal/api/api_test.go
+++ b/sherlock/internal/api/api_test.go
@@ -1,0 +1,32 @@
+package api
+
+import (
+	"embed"
+	"github.com/stretchr/testify/assert"
+	"path"
+	"regexp"
+	"testing"
+)
+
+//go:embed *
+var apiFiles embed.FS
+var capsInApiPathRegex = regexp.MustCompile("/api/\\w*[A-Z]")
+
+func TestApiFiles(t *testing.T) {
+	validateApiPathsInDirectory(t, ".")
+}
+
+func validateApiPathsInDirectory(t *testing.T, subdirectory string) {
+	entries, err := apiFiles.ReadDir(subdirectory)
+	assert.NoError(t, err)
+	for _, entry := range entries {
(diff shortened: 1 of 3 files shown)


choover-broad (approved):
approved! asking 'cause I'm curious - it looks like the test examines .go source filenames under `sherlock/internal/api`, but the typo was in a comment. Does the test somehow validate comments for correctness?

jack-r-warren:
> approved! asking 'cause I'm curious - it looks like the test examines .go source filenames under `sherlock/internal/api`, but the typo was in a comment. Does the test somehow validate comments for correctness?

@choover-broad Yeah, it's looping over file/folder names just so it can handle nested directory stuff but it's reading in file contents to the `data` variable [here](https://github.com/broadinstitute/sherlock/pull/327/files#diff-8f84f880e0af871d05ef0c0cc01645c8c439625e105293c40a1aab82c59055b9R27) and running the regex against that [here](https://github.com/broadinstitute/sherlock/pull/327/files#diff-8f84f880e0af871d05ef0c0cc01645c8c439625e105293c40a1aab82c59055b9R29)

PR #335, opened 2023-10-18
DDO-3242: Remove "v2_" from table names

katiewelch:
When we migration from sherlock v1 to v2, we renamed many database tables, indexes, foreign keys, keys, and sequences to include "v2_" in the name. Now, we are migrating to sherlock v3 and these resources no longer accurately reflect the current environment. During the first migration to v2, we were creating new resources. In this migration to v3,  we will still be using the same resources, and will continue to  use these resources for the foreseeable future. We will not add "v3_" to the resource names, so that if we do further migrations in the future, these resources will continue to reflect the current state.

This change included two parts.
1. Find and replace all references to "v2_" in the Sherlock database.
2. Database migrations. The migration was a simple `rename` command in both the up and down migrations. Unfortunately, `rename` commands cannot be stacked, so each rename had to be written as it's own command. The migrations are organized by table and then split into keys, foreign keys, and indexes. All sequence commands are at the bottom.

### Testing
I tested this locally by
1. starting up my local Sherlock. Doing this will run all the migrations and fail if there are any errors.
Screenshot 2023-10-24 at 12 09 31 PM
2. using the [golang-migrate/migrate](https://github.com/golang-migrate/migrate) tool to test run my up and down migrations. I used 
`migrate -path sherlock/db/migrations -database 'postgres://sherlock:password@localhost:5432/sherlock?sslmode=disable' goto 59` 
and 
`migrate -path sherlock/db/migrations -database 'postgres://sherlock:password@localhost:5432/sherlock?sslmode=disable' goto 60` 
to test down and up, respectively.
Screenshot 2023-10-24 at 12 12 39 PM

### Risk
This change will affect many resources in the database and will require downtime,  but is easy to revert using ArgoCD. We will push to dev first and make sure it spins up.


jack-r-warren (approved):
Looks great! My one comment is that I don't think we should change metrics/metrics.go in this PR because that'll impact Prometheus, not the DB, and we'll want to roll that out differently than the database changes

PR #338, opened 2023-10-27
[DDO-3268] Improve logging/metrics for Gin responses

jack-r-warren:
Adds metrics for response latency/count, by Gin's matched route.

Also touches up logging. Normal internal log messages are staying the same, but the Gin ones now take full advantage of Zerolog's JSON fields. Things like the status code, client IP, latency, etc are no longer logged in just a big string, they're logged as fields with Zerolog.

This makes the Gin logging slightly worse locally, but _way, way_ better through GCP. It's a balancing act and I think the Gin logs are primarily useful through GCP anyway, so I think it's okay. This also helps avoid logging things to console that we only need on GCP, like the client IP.

Before:
Screenshot 2023-10-27 at 3 32 40 PM
After: 
Screenshot 2023-10-27 at 3 52 42 PM

## Testing

Screenshots above; visited /metrics to validate that part.

## Risk

Very low, no functionality changing, just logs

### DIFF
--- a/sherlock/internal/config/config.go
+++ b/sherlock/internal/config/config.go
@@ -86,6 +86,7 @@ func configureLogging(infoMessages ...string) {
 	if Config.String("mode") == "debug" {
 		// Colored text for CLI
 		output := zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.Kitchen}
+		// If the call-site of the log message is included, format it nicely for human consumption
 		output.FormatCaller = func(i interface{}) string {
 			if i == nil {
 				return ""
--- a/sherlock/internal/config/config.go
+++ b/sherlock/internal/config/config.go
@@ -13,6 +13,7 @@ import (
 	stdlog "log"
 	"os"
 	"strings"
+	"time"
 )
 
 // Config holds Sherlock's global configuration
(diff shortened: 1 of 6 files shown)


sherlock/internal/config/config.go
```diff
@@ -84,16 +85,24 @@ func configureLogging(infoMessages ...string) {
...
 		// Colored text for CLI
-		logBuilder = log.Output(zerolog.ConsoleWriter{Out: os.Stderr}).With()
+		output := zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.Kitchen}
+		output.FormatCaller = func(i interface{}) string {
+			if i == nil {
+				return ""
+			} else {
+				return fmt.Sprintf("%-50s |", strings.TrimPrefix(fmt.Sprintf("%s", i), "/build/sherlock/internal"))
+			}
+		}
```
mflinn-broad, line 95: Curious what this is for?

mflinn-broad (approved):
One question, otherwise looks straightforward to me!

sherlock/internal/config/config.go
jack-r-warren: You can set `log.caller: true` in the config and then this function formats out the call-site of the log message:

Screenshot 2023-10-27 at 4 09 26 PM

Added a comment to the function to explain that in-line

PR #356, opened 2023-11-07
DDO-3269: Environment V3 SQL Migrations

katiewelch:
We want to add the environment type to sherlock v3. This PR is doing the database migrations for that table based off of [this function](https://github.com/broadinstitute/sherlock/blob/14430c6af228994fc1e8b11c9a79259b4465ea4b/sherlock/internal/deprecated_models/v2models/environment.go#L164) from the deprecated model.

Testing
I tested this locally by

1) starting up my local Sherlock. Doing this will run all the migrations and fail if there are any errors.
Screenshot 2023-11-07 at 11 28 03 AM

2) using the [golang-migrate/migrate](https://github.com/golang-migrate/migrate) tool to test run my up and down migrations. I used
`migrate -path sherlock/db/migrations -database 'postgres://sherlock:password@localhost:5432/sherlock?sslmode=disable' goto 61`
and
`migrate -path sherlock/db/migrations -database 'postgres://sherlock:password@localhost:5432/sherlock?sslmode=disable' goto 62`
to test down and up, respectively.
Screenshot 2023-11-07 at 11 28 55 AM


sherlock/db/migrations/000062_environment_v3.up.sql
```diff
@@ -0,0 +1,54 @@
...
+        check ((lifecycle = 'template' and template_environment_id is null) or
+               (lifecycle = 'dynamic' and template_environment_id is not null) or
+               (lifecycle = 'static' and
+                base != '' and
+                default_cluster_id is not null and
+                requires_suitability is not null));
+
+alter table environments
+    add constraint default_namespace_present
+        check (default_namespace != '');
```
jack-r-warren, line 21: unfortunately I think SQL null passes this check. I know we probably wouldn't be able to insert null through Gorm in most cases, but I think we should write these sorts of checks (here, line 29, line 15) like you did on line 25. I don't think you need to change the go tests at all because yeah you'd have a hard time inserting null

sherlock/db/migrations/000062_environment_v3.up.sql
```diff
@@ -0,0 +1,54 @@
...
+               (legacy_owner is not null and legacy_owner != ''));
+
+alter table environments
+    add constraint lifecycle_valid
+        check ((lifecycle = 'template' and template_environment_id is null) or
+               (lifecycle = 'dynamic' and template_environment_id is not null) or
+               (lifecycle = 'static' and
+                base != '' and
+                default_cluster_id is not null and
+                requires_suitability is not null));
```
jack-r-warren, line 17: I'm not sure this is quite right -- https://github.com/broadinstitute/sherlock/blob/main/sherlock/internal/deprecated_models/v2models/environment.go#L189 has all of the static checks also apply to dynamic. Template gets to break some rules because we won't actually end up creating it in our infrastructure, but both dynamic and static need to have base/default cluster/requires suitability

jack-r-warren (approved):
LGTM!

PR #395, opened 2023-12-13
DDO-3377: add payload to error

katiewelch:


sherlock-webhook-proxy/pkg/handler.go
```diff
@@ -255,9 +256,15 @@ func HandleWebhook(w http.ResponseWriter, r *http.Request) {
 			})
 
 			// Handle response cases
+			payloadPretty, err2 := json.MarshalIndent(payload, "", "    ")
+			payloadPrettyString := ""
```
jack-r-warren, line 260:
```suggestion
			var payloadPrettyString string
```
**nitpick (non-blocking):** no need to assign to empty (I think these will compile identically but it's just a style thing)

jack-r-warren (approved):

PR #398, opened 2023-12-14
DDO-3377: Use CreatedAt instead of CommittedAt

katiewelch:


sherlock/internal/api/sherlock/git_commits_v3_upsert.go
```diff
@@ -8,15 +8,14 @@ import (
...
-	"time"
 )
 
 type GitCommitV3Upsert struct {
-	GitRepo      string    `json:"gitRepo"`
-	GitCommit    string    `json:"gitCommit"`
-	GitBranch    string    `json:"gitBranch"`
-	IsMainBranch bool      `json:"isMainBranch"`
-	CommittedAt  time.Time `json:"committedAt"`
+	CommonFields
```
jack-r-warren, line 14:
```suggestion
```
**suggestion:** I don't think we actually want to allow upserting the CommonFields, those are typically ready only (the underlying `gorm.Model` is handled by Gorm, so I actually don't know what would happen if you tried to manually set these)

sherlock/internal/api/sherlock/git_commits_v3_upsert.go
```diff
@@ -54,15 +53,14 @@ func gitCommitsV3Upsert(ctx *gin.Context) {
 	var timeSince *uint
 
 	if len(previous) > 0 {
-		timeSince = utils.PointerTo(uint(body.CommittedAt.Sub(previous[0].CommittedAt).Seconds()))
+		timeSince = utils.PointerTo(uint(body.CreatedAt.Sub(previous[0].CreatedAt).Seconds()))
```
jack-r-warren, line 56: **suggestion:** Based on my comment above, the body won't have a `.CreatedAt`. Instead I think we should just use `time.Now()` (not leaving it as a GitHub suggestion because you'd need to import `"time"` too, which GoLand will do if you write it yourself)

jack-r-warren (commented):
**thought:** This PR does two things that normally would not be okay, but I think it is okay here:

One, the cloud function and the backend are being changed by the same PR, but the cloud function will get immediately auto-deployed (in practice, before we would've deployed the new Sherlock). This means Sherlock before this PR would receive some incoming git-commit requests lacking the CommittedAt field, which'll make it unhappy.

Two, this PR drops a column while it is still in use (both the old Sherlock and the new Sherlock will coexist briefly during an update). Old Sherlock would get very confused and unhappy if it tries to read or write the CommittedAt column if new Sherlock has already migrated and dropped it.

But... this table and everything here is completely isolated and we're going to throw out the data anyway, we're just moving fast to collect data. I don't think we should worry about those two notes I just made; who cares if we get an ugly row or two.

jack-r-warren (approved):

PR #413, opened 2024-01-04
[DDO-3400] Bump go version to 1.21

jack-r-warren:
My Slack message code uses parts of the standard library that aren't in 1.19, so let's do an upgrade :/ this'll probably ripple into Thelma but I'm not going to worry about that just yet lol

Requires bumping the alpine version used to build Sherlock from 3.16 to 3.19, but that's just the build environment, so it seems fine

## Testing

Tests pass, if there's an issue it'll pop up on dev

## Risk

Low? I wouldn't expect runtime issues from this sort of thing


sherlock/internal/models/test_data_test.go
```diff
@@ -15,7 +14,6 @@ func (s *modelSuite) TestBulkData() {
 	// Rather than iterating through the methods in order, use rand.Perm to effectively
 	// shuffle the order we process each index.
 	// https://pkg.go.dev/math/rand#Rand.Perm
-	rand.Seed(time.Now().UnixNano())
```
em-may, line 18: Is this not needed anymore on the new version?

em-may (approved):
Left a comment

sherlock/internal/models/test_data_test.go
jack-r-warren: Correct! Got a linter error actually that prompted me to make this change, see [here](https://github.com/broadinstitute/sherlock/actions/runs/7412115470/job/20168179700?pr=413)

PR #416, opened 2024-01-04
[DDO-3400] Wrap deployment notification blocks with sections

jack-r-warren:
Go's type system isn't wonderful and that's sorta reflected in this library, which seems to do some plainly not allowed things if you ask it to


em-may (approved):
Ugh block kit, my nemesis.

PR #460, opened 2024-01-30
[DDO-3449] Database Instance, v3 methods

jack-r-warren:
create, edit, list, get, and delete for v3 database instances

also fixed some missing duplicate validation and fixed a bug with duplication that it revealed

## Tests

New file for each route and for the data type itself

## Risk

Very low


choover-broad (approved):
thanks for this 🙇

PR #480, opened 2024-02-12
[DDO-3470] Remove deprecated fields from v3 APIs

jack-r-warren:
See the API doc comment for the specific updates. Removes fields from the v3 APIs/models in line with the Slack discussions from today. v2 will continue to work as it does currently.

This PR also removes the metric that tracked how many charts used firecloud-develop

## Testing

Source and tests updated

## Risk

Low


jyang-broad (approved):
appears to remove fc-dev refs from v3 apis, also updates v3 tests, doesn't touch v2.

PR #481, opened 2024-02-13
[DDO-3470] Fix Changeset response when there were no changes

jack-r-warren:
Discussed in https://broadinstitute.slack.com/archives/CQ6SL4N5T/p1707836110299439, the issue is that `db.Find(&ret, []uint{})` loads everything rather than nothing. So we guard that and bail out beforehand.

## Testing

Added tests for both /plan and /plan-and-apply for exactly this case (in two different variants)

## Risk

Low


em-may (approved):
Nice find

PR #486, opened 2024-02-14
[DDO-3470] Add GitHub info to request logs when present

jack-r-warren:
The kinds of madness we can do with Sherlock's authentication system. Six line diff including comments to log the workflow that made a given call to Sherlock.

I'm doing this to try to figure out what on earth is still calling Sherlock v2 endpoints.

## Testing

Worked locally. Deployed to sherlock-dev and used the GitHub auth test action to issue a call to it from GHA, this showed up in the logs! Truncating everything else so I don't post IPs and stuff.

Screenshot 2024-02-14 at 3 04 20 PM

## Risk

Very low


jyang-broad (approved):
:stamp:

PR #518, opened 2024-03-11
[DDO-3535] Actually delete deleted files during client library generation

jack-r-warren:
Systemically fix the issue surfaced from https://github.com/broadinstitute/sherlock/actions/runs/8235205030/job/22518779627, where "deleted" generated client library files wouldn't get properly deleted and could cause problems down the line if they expose models that shouldn't exist anymore of if the files just get out of date and don't compile.

## Testing

Did this process locally and it seemed to fix the issue seen in CI but we'll see

## Risk

Very low, either the build works or it doesn't

### DIFF
--- a/.github/workflows/sherlock-build.yaml
+++ b/.github/workflows/sherlock-build.yaml
@@ -129,6 +129,11 @@ jobs:
       - name: Update Pact provider documentation
         run: make document-pact-provider
 
+      - name: Delete existing Go client library code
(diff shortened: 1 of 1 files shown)


jyang-broad (approved):
does what the label says it does, deletes the client. maybe add comments in line for reasoning as it's not entirely obvious to the untrained eye why you're manually deleting the client before a new build.

PR #519, opened 2024-03-13
[DDO-3385] Add test coverage for Sherlock-Pact code

ichengchang:
**Objective** The objective is to add test coverage for the `changesetPostApplyActions` function to ensure that the "Report to Pact" code never panics.

**Parameters Required for `packbroker.RecordDeployment`** The `packbroker.RecordDeployment` API call within the "Report to Pact" step requires three parameters: 

- chart.Name
- appliedChangeset.To.AppVersionExact
- environment.PactIdentifier.

**Test Plan**

Step 1: The initial test will focus on asserting that the step never panics when `appliedChangeset.To.AppVersionExact` is set to nil.
Step 2 (Optional): If necessary, additional tests can be added to handle scenarios where `environment.PactIdentifier` is missing.
Step 3 (Optional): Further, a test case can be added to spin up a fake server locally to mock the `RecordDeployment` route, allowing for more comprehensive testing.


knapii-developments (approved):

sherlock/internal/models/changeset_plan_apply.go
```diff
@@ -204,18 +205,18 @@ func changesetPostApplyActions(db *gorm.DB, appliedChangesets []Changeset) {
...
-				//		go slack.ReportError(context.Background(), fmt.Sprintf("unable to load chart %d after applying changeset %d", appliedChangeset.ChartRelease.ChartID, appliedChangeset.ID), errToSwallow)
-				//	} else if chart.PactParticipant != nil && *chart.PactParticipant && appliedChangeset.To.AppVersionExact != nil {
-				//		go pactbroker.RecordDeployment(
-				//			chart.Name,
-				//			*appliedChangeset.To.AppVersionExact,
-				//			*environment.PactIdentifier)
-				//	}
-				//}
+				if environment.PactIdentifier != nil {
+					var chart Chart
```
jyang-broad, line 209: This is a lot of pact specific logic in this method, 
some wrapper method like pactbroker.ReportChangesetBump (or something) should just get called here and forgotten about

ichengchang: Let's do this in another ticket as the focus of this ticket is to get the pactbroker test to prove there is no panic when a changeset is applied.

araina2018:
So step1 and 2 is done. Does DevOps think we still need Step2 or that's covered now?

PR #521, opened 2024-03-26
[DDO-3577] API for GHA jobs

jack-r-warren:
Adds a simple API to store GHA jobs with data on when they were queued, when they started, and when they completed. I'm not adding automated testing here, at least not right now -- Sherlock's tests will check that this won't break anything else, and I just want to move as quickly as possible to start collecting data.


broadbot (approved):

em-may (approved):
I am (not) very good at switching accounts...

PR #542, opened 2024-04-30
/api/environments/v3 [put] (sketch)

jack-r-warren:
**This _doesn't_ do any version changes.** It's a normal CRUD endpoint for upsert, a `PUT` at `/api/environments/v3`. It supports how I'm guessing pooling would be implemented, but there's a long comment on the one line we might need to change if I'm wrong.


jack-r-warren:
Closing due to stale

PR #590, opened 2024-06-28
[DDO-3763] Do CORS from Sherlock

jack-r-warren:
[Details in the ticket.](https://broadworkbench.atlassian.net/browse/DDO-3763)

This does the CORS thing that we've been having [the proxy do so far.](https://github.com/broadinstitute/terra-helmfile/blob/master/charts/sherlock/templates/configMap.yaml#L34-L40) Benefit now is that we can actually just pass the list of origins and it filters based on them, without treating them like a vulnerable regex like Apache does. Argument for this approach is that it's simpler.

We also move the CSRF mitigations from the proxy into Sherlock for the same reasons.

Two chart changes are required for this to work: we'll need to pass the origins into Sherlock's config and we'll need to configure the proxy to not mutate the Host/Origin headers so that this new middleware can see the real values.

## Testing

Via unit test, my own manual testing, and most importantly, Sarah Gibson's testing from her pentest tools

## Risk

Low

### DIFF
--- a/sherlock/config/default_config.yaml
+++ b/sherlock/config/default_config.yaml
@@ -20,6 +20,9 @@ log:
   caller: false
   level: debug
 
+cors:
+  allowOrigins: []
+
 db:
   # Can be "pgx" or "cloudsql-postgres". "cloudsql-postgres" won't work in debug mode.
   driver: pgx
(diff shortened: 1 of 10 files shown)


mflinn-broad (approved):
Makes sense to me

choover-broad (approved):

PR #591, opened 2024-06-28
[DDO-3765] Absorb Google Workspace Directory API Quota Issues

jack-r-warren:
See https://broadinstitute.slack.com/archives/CADM7MZ35/p1719608453224929

Sherlock will detect the dailyLimitExceeded error string, log the error, but move on, allowing Sherlock to boot and operate normally. It will continue to rely on it's cache of suitability data that it stores in the database


choover-broad (approved):

jyang-broad (commented):
swallows an expected error.

PR #608, opened 2024-07-15
[DDO-3785] OIDC

jack-r-warren:
- Implements OIDC for ["authorization service" usage](https://docs.google.com/document/d/1yF49f6Twte5U7eyKuY2tI2PiaKB1tXzNw23o5aJH6mU/edit#heading=h.ig4gkffyaaxu), with `/oidc` as the issuer and `/login` as the "login" page that just validates auth requests based on middleware
- Adds configuration option to have Sherlock never report an empty/null RequiredRole and instead default to a static string. This is a request from appsec as its easier to have downstream consumers use it verbatim than say that they properly implement some logic like `environment.requiredRole ?? "all-users"`

## Testing

I've added tests for our implementation of the OIDC library we're using and the login page. Appsec has tested this in concert with ArgoCD and it got the thumb from them.

## Risk

Low at this point due to testing.


jyang-broad (approved):

sherlock/internal/api/login/handler_test.go
```diff
@@ -0,0 +1,37 @@
+package login
```
mflinn-broad, line 1: What is this testing?

sherlock/internal/oidc_models/auth_request.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/broadinstitute/sherlock/go-shared/pkg/utils"
+	"github.com/broadinstitute/sherlock/sherlock/internal/models"
+	"github.com/google/uuid"
+	"github.com/zitadel/oidc/v3/pkg/oidc"
+	"github.com/zitadel/oidc/v3/pkg/op"
+	"time"
+)
+
+// "AuthRequest implements the op.AuthRequest interface"
+var _ op.AuthRequest = &AuthRequest{}
```
mflinn-broad, line 14: I like this pattern is an easily visible compile time validation

mflinn-broad (approved):
Non blocking comments, other wise cool stuff.

sherlock/internal/api/login/handler_test.go
jack-r-warren: handler_test is sorta borrowing the convention from the other (more multi-endpoint) api packages -- this hosts the suite and the other test files (login_test.go here) use it.

sherlock/internal/oidc_models/auth_request.go
jack-r-warren: Yeah it's convenient because if you write this as the first step GoLand will offer to generate stubs for all the methods

PR #617, opened 2024-07-31
[DDO-3785] Ignore suitability within Sherlock

jack-r-warren:
I'll remove it from the API and then the database at a later junction. Requires all other PRs for DDO-3785 to be merged first.

## Testing

Updated tests to all reference roles rather than suitability. Had more than I thought, I thought I updated tests before, but I'm not arguing.

## Risk

Low


jyang-broad (approved):
removes "suitability" as a top-level attribute, now lives somewhere in roles.

mflinn-broad (approved):

PR #622, opened 2024-08-06
[DDO-3829] Document host on Swagger page

jack-r-warren:
Making a fully valid Swagger spec from Sherlock's dynamic doc.json endpoint helps appsec's tools run against it. We'll need to specify the primaryHost from helmfile. Can't borrow origins here because origins is a list for use with CORS, and those have schemes anyway.

Also adds `/` as a base path. I've confirmed on the Swagger page that there's no difference in behavior here. Since doc.json sets basepath to an empty string if you don't provide it, that too was failing some validation.

Setting basepath to `/` is the same as what should be done automatically: https://swagger.io/docs/specification/2-0/api-host-and-base-path/

It's okay for this to ship immediately because the empty primaryHost value is the same behavior as we've got right now.

## Testing

I used the Swagger page generated on this PR and confirmed that there is no difference in behavior. I don't expect this to cause any downstream issues. If there is an issue, downstream tests would catch it before a broken usage of the Swagger spec got promoted. The API here isn't changing.

## Risk

Low.

### DIFF
--- a/sherlock/config/default_config.yaml
+++ b/sherlock/config/default_config.yaml
@@ -5,6 +5,12 @@
 #   It will use mock authentication and authorization.
 mode: debug
 
+# The primary host name that Sherlock should document itself as running on, used in the
+# Swagger docs hosted dynamically. The Swagger page will work with this empty, but other
+# tools (like AppSec's scanners) that need a fully valid config will fail.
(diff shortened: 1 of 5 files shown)


sarahgibs (approved):
Looks good. Thanks!

mflinn-broad (approved):

PR #643, opened 2024-08-23
[DDO-3857] Firecloud account manager: ignore suspended users

jack-r-warren:
An impressive thing to forget when writing this code but whatever. Before it would've just repeatedly kept suspending people which would've flooded us with notifications (as in, it would've sent one ever hour).

## Testing

Modified test cases

## Risk

Very low (security impact non-zero but we haven't even turned this on, and again, the actual functionality would've been okay, it just would've notified us too much)


jyang-broad (approved):
:stamp:

em-may (approved):

PR #648, opened 2024-08-27
[DDO-3859] Allow domain-wide delegation from Sherlock's Firecloud account manager

jack-r-warren:
I've discovered an annoying feature of Google Workspace: a service account, even with admin permissions, seems unable to use those permissions against super admins. This defeats the point of Sherlock being able to suspend even an admin's account.

The workaround is to use domain-wide-delegation and impersonate an admin to use the API. In Go, this requires using an entire separate package (and specify a bunch of technically-unnecessary fields that are required by the validation but only necessary for service account impersonation).

As part of getting this to work I had to do a big giant dependency update to fix an issue with mismatched Google client library versions.

## Testing

DWD is one of those things that's really, really hard to fit into any automated test. It's just the magic incantation to call the client library.

So I deployed this build to dev and told it to run the Firecloud account manager against me, Katie, and Lucy in the test Firecloud (with a window so short it would think it would need to ban the two of them). It ran without issue (which means it was able to ban Katie, a super admin).

## Risk

Low in my book. This changes how Sherlock authenticates to Firecloud.org for this specific functionality, but Sherlock has access to these same scopes on its own. This is just a change to how the client library is instantiated. Any failures here are surfaced in notifications but don't impact the rest of Sherlock's operations.


sherlock/config/default_config.yaml
```diff
@@ -474,7 +480,9 @@ firecloudAccountManager:
   #     - "user-to-never-disable@test.firecloud.org"
   #   # How long to allow a new account to exist without a login before it is disabled. We set this low
   #   # to help make sure the invite emails Google Workspace sends aren't a security risk.
+  #   # The largest unit that can be passed here is hours (so it's unambiguous about leap sends etc.)
```
em-may, line 483:
```suggestion
  #   # The largest unit that can be passed here is hours (so it's unambiguous about leap secs etc.)
```
Wondering if this is a typo, not sure what 'leap sends' are 🤔

sherlock/config/default_config.yaml
```diff
@@ -474,7 +480,9 @@ firecloudAccountManager:
...
   #   newAccountGracePeriod: 24h
   #   # How long an account can go without a login before being disabled.
-  #   inactivityThreshold: 90d
+  #   # The largest unit that can be passed here is hours (so it's unambiguous about leap sends etc.)
```
em-may, line 486: Same as above.

em-may (approved):

katiewelch (approved):

sherlock/config/default_config.yaml
jack-r-warren: Good find, thanks

PR #661, opened 2024-09-10
[DDO-3870] Azure account provisioning via propagation #minor

jack-r-warren:
Ports [broadinstitute/terra-privileged-user-sync](https://github.com/broadinstitute/terra-privileged-user-sync) into Sherlock.

The cronjob is responsible for basically four steps. In Sherlock each one is a separate propagation engine.

1. Create the account in the non-B2C tenant -- new [AzureAccountEngine](https://github.com/broadinstitute/sherlock/pull/661/files#diff-4928b53d89e6818d126be0a154f3c531681c7060a9202016106a30596d257751)
2. Add the account to a permission group in the non-B2C tenant -- existing [AzureGroupEngine](https://github.com/broadinstitute/sherlock/blob/main/sherlock/internal/role_propagation/propagation_engines/azure_group.go)
    - We've already replaced this by rolling out the break-glass capability
3. Invite the account to the B2C tenant -- new [AzureInvitedAccountEngine](https://github.com/broadinstitute/sherlock/pull/661/files#diff-eac65d499288d6a929b3990b89df8ebcfa42001666c30d4ea864d896c6e8b531)
    - Rather than having an account per-tenant, this is how we accomplish account per-environment
4. Give the account an admin role in the B2C tenant -- new [AzureDirectoryRoleEngine](https://github.com/broadinstitute/sherlock/pull/661/files#diff-7971b0789277d805683a418514d6a0f46990ab3dbb781eb8aa400701e87921a2)
    - Permission groups won't work in B2C so we have to grant the role directly

Splitting this across engines creates a few differences in behavior:
- The existing cronjob has some logic to handle propagation delay on Azure's end between the steps. Sherlock will run them sequentially to be optimistic (see [role_propagation's Init](https://github.com/broadinstitute/sherlock/pull/661/files#diff-4f0a5dfcd1b1d52da9da284eebc94d6adac277948dc9f85b59e0874f9b90f900)) but it doesn't explicitly handle cloud provider delay. This is okay because each engine instance does independent drift correction, so the state will be eventually consistent even in cases of extreme delay on Azure's end.
- The existing cronjob uses Google Workspace as the source of truth. Sherlock uses itself.
- The existing cronjob handles account suspension but not changes to other account fields like display name. Sherlock will manipulate these fields based on itself.

Despite using propagation engines, accounts will not be deleted when a Sherlock RoleAssignment is removed. Instead, the necessary engines will merely suspend the user and tolerate that going forward (so it won't keep trying to remove the same users over and over again).

Propagation also now supports per-propagator-instance dry-run, where mutation operations on the engine will be logged rather than called. This will help us roll out this capability while acceptably matching the existing cronjob's behavior.

## Testing

Tests added where possible, but for propagation engines the actual behavior of the client libraries is something we only truly validate when running against the cloud. See below.

## Risk

The risk in this PR comes from two places:
1. The Azure client library's behavior has some unknowns around it. It's possible that propagation could fail (with no impact to other engines, but still)
2. The behavior changes in Sherlock's implementation could cause some shift in accounts -- since Sherlock will be commandeering something managed by another tool, there may be a noticeable diff

Point 2 is heavily mitigated by the dry-run feature, but point 1 is complicated by the fact that we don't have a destructible set of Azure tenants to let this feature loose on. 

If we enable this feature in dev Sherlock pointed just at dev Azure... Sherlock's own source of truth in dev is miles away from what the cronjob is used to using. The diff will be large and unrealistic. It will want to ban a bunch of actual accounts that people use on a day-to-day basis (just because no one uses Sherlock dev). We can dry-run, but there's a last mile that I think we may choose not to enable from Sherlock dev.

I think this may be an instance where it makes sense to rely on Sherlock's error recovery and engine segmentation to just... turn this on in prod. More specifically, my proposal:

1. Sherlock dev dry-run against Azure dev
2. Sherlock prod dry-run against Azure dev
3. Sherlock prod run against Azure dev
4. Sherlock prod dry-run against Azure prod
5. Sherlock prod run against Azure prod

In other words, there's a `AzureAccountEngine.Remove(user)` method that I think we may not want to actually run from Sherlock dev -- because Sherlock dev will want to do it in an unrealistic and destructive way (because it, by nature, doesn't have prod's knowledge of DSPers). I think it's low risk for the first execution of that to be in prod, because even if it fails, none of the rest of Sherlock would be impacted.


choover-broad (approved):

jyang-broad (approved):
:stamp:

PR #700, opened 2024-10-31
A demo/draft of spreadsheet code for asset management

jack-r-warren:
I put a test spreadsheet in the shared drive: https://docs.google.com/spreadsheets/d/1isOybP49rb6n7uX2vtT_azKj-sgRzp7RWkOHxFSsSKE/edit?gid=591394540#gid=591394540

Idea is you paste that link in and it'll do the thing. For the "unique resource identifier", that'll need to be the full resource name you get from the cloud asset API. Sarah T said that because that full name is long and verbose, it'd be great to also have the tail end of it put in a separate short asset name column.

You'll need to set up ADC in a specific way for this to work.

```bash
gcloud auth application-default login --scopes=https://www.googleapis.com/auth/userinfo.email,https://www.googleapis.com/auth/accounts.reauth,https://www.googleapis.com/auth/userinfo.profile,https://www.googleapis.com/auth/sqlservice.login,openid,https://www.googleapis.com/auth/cloud-platform,https://www.googleapis.com/auth/calendar,https://www.googleapis.com/auth/spreadsheets
```

The calendar scope is helpful for Beehive's read-workbench-protection-calendar feature, while the spreadsheet scope is what powers this read-and-mutate-spreadsheet thingy. I don't think we'll need to explicitly list those scopes for Kubernetes or anything but you will need to enable the API in the project.

That cloud asset API is here https://pkg.go.dev/cloud.google.com/go/asset/apiv1. https://pkg.go.dev/cloud.google.com/go/asset/apiv1#Client.QueryAssets is probably what you want.

### DIFF
--- a/sherlock/config/default_config.yaml
+++ b/sherlock/config/default_config.yaml
@@ -662,3 +662,7 @@ firecloudAccountManager:
   #   # The largest unit that can be passed here is hours (so it's unambiguous about leap secs etc.)
   #   inactivityThreshold: 2160h
   # ```
+
+assetInventory:
+  enable: true
+  requiredRole: ""
--- a/sherlock/docs/docs.go
+++ b/sherlock/docs/docs.go
@@ -433,6 +433,76 @@ const docTemplate = `{
                 }
             }
         },
+        "/api/asset-inventory/procedures/v3/generate": {
+            "post": {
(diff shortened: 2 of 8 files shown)


jack-r-warren:
Closing due to stale

PR #713, opened 2024-11-19
[DDO-3900] Add service banner bucket field to environments

em-may:
# What
Nicely outlined by Jack in [DDO-3900](https://broadworkbench.atlassian.net/browse/DDO-3900).

TL;DR: Adds a service banner bucket field to environments to allow setting where their Terra UI banner bucket is so we can add more features to set/get/clear the bucket in Sherlock/Beehive later on, replacing the Jenkins banner set/clear jobs, which will allow us to deprecate/shut down Jenkins.

## Details
All environments will have a `service_banner_bucket` field in the db, which can be be `null` or a string representing the name of the GCS bucket where the Terra UI banner files live. If a dynamic environment (BEE) is created from a template, one of the following things will happen:

* If the BEE's template has a service banner bucket set, the BEE will inherit it. I plan to use subdirs within buckets to allow for per-BEE banner files within the same bucket.
* If the BEE's template does not have a service banner bucket set, the BEE won't have a service banner bucket set.

You can change the banner bucket at any time, currently only with the APIs directly, but eventually Beehive will have a metadata field where you will be able to edit it.

# Testing
More tests included in this PR cover the model and API functionality. I also manually did the following to test the new field and functionality:
1. Spin up Sherlock + Postgres locally
2. Create a BEE cluster
3. Create a swatomation template with a default service banner bucket set
4. Create a swatomation template without a default service banner bucket set
5. Create a BEE from (4), the service banner bucket is set, as expected
6. Create a BEE from (5), the service banner bucket is not set, as expected

[DDO-3900]: https://broadworkbench.atlassian.net/browse/DDO-3900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sherlock/db/migrations/000102_add_banner_field_to_environments.down.sql
+++ b/sherlock/db/migrations/000102_add_banner_field_to_environments.down.sql
@@ -1,0 +1,2 @@
+alter table environments
+    drop column if exists service_banner_bucket;
--- a/sherlock/db/migrations/000102_add_banner_field_to_environments.up.sql
+++ b/sherlock/db/migrations/000102_add_banner_field_to_environments.up.sql
@@ -1,0 +1,2 @@
+alter table environments
+    add if not exists service_banner_bucket text;
--- a/sherlock/docs/docs.go
+++ b/sherlock/docs/docs.go
@@ -5069,6 +5069,11 @@ const docTemplate = `{
                     },
                     {
                         "type": "string",
+                        "name": "serviceBannerBucket",
+                        "in": "query"
+                    },
+                    {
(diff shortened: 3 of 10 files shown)


choover-broad (approved):
LGTM. Nice work!

jyang-broad (approved):
Adds new text column to `environments`

not your problem, but still a nit: single letter e for Environment? I was confused for a bit thinking e == error.

PR #714, opened 2024-11-20
[DDO-3900] Fix clearing banner bucket

em-may:
In #713 I missed an edge case, where trying to clear the banner bucket sets the db value to the empty string instead of null.

# Testing
Tests added for this case and I verified with local Sherlock + Beehive.


choover-broad (approved):
nice find!

jyang-broad (approved):

